PR #1026, opened 2026-07-21
feat: #1024 1024-part-cockpit-remote-gates

christrudelpw:
## Summary

Closes #1024

---


packages/orchestrator/src/__tests__/cockpit-gates-integration.integration.test.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  // ---------------------------------------------------------------------------
+  // Scenario S1a — Gate open → cluster.cockpit event (FR-003).
+  //
+  // TODO(#1020, #1021): unskip once the gates module exports
+  // `gateOpenFixture` (S-10) via `@generacy-ai/cockpit/gates` (S-9) and the
+  // orchestrator adds `POST /cockpit/gates` + `'cluster.cockpit'` to
+  // ALLOWED_CHANNELS (S-2, S-4).
+  // ---------------------------------------------------------------------------
+  it.skip('S1a — gate open POST produces a cluster.cockpit event byte-equal to the contract', async () => {
```
christrudelpw: **Suite passes vacuously — every scenario is `it.skip()` with its body commented out.** All eight scenarios (S1a, S1b, S2, S3, S4, S5, F1, F2, F3) are skipped sketches with no executable assertions and no fixture imports. The integration job (`vitest.integration.config.ts`, `passWithNoTests: true`) reports green even if #1021's `POST /cockpit/gates` emits a wrong-shaped `cluster.cockpit` event, the answers-file writer never dedups, or the doorbell never emits a `gate-answer`. #1024's acceptance criteria ("harness runs green using the shared contract fixtures", "prove the whole cluster-side path") are not exercised — please implement (unskip) the scenarios with real assertions against the shared fixtures.

packages/orchestrator/src/__tests__/cockpit-gates/scenario-helpers.ts
```diff
@@ -0,0 +1,170 @@
...
+      extraArgs: opts.doorbellArgs ?? [],
+    });
+  } else {
+    doorbell = createDoorbellDriver({
+      answersFilePath,
+      env: { COCKPIT_ANSWERS_FILE: answersFilePath },
+      extraArgs: opts.doorbellArgs ?? [],
+    });
+    // TODO(#1023): uncomment once the doorbell tails an answers file:
+    //   await doorbell.start();
```
christrudelpw: **Harness infrastructure is never actually run.** `setupScenario()` constructs the doorbell driver but never `start()`s it (the call is commented), orchestrator boot is a hardcoded `null` placeholder (~line 85), and `awaitCockpitEvents()` throws unconditionally (~line 165). Clarification Q3 requires the doorbell to run as a **real spawned child** so FR-007 kill/restart is genuine, and S2/S5 assert the `cockpit_await_events` typed batch — at runtime neither happens. Because the fake-peer WS flow, doorbell spawn/restart, and api_request correlation are never executed, latent bugs there won't surface here. Please wire `setupScenario()` to boot the orchestrator, spawn the real doorbell child, and drain the MCP bus.

christrudelpw (commented):
**Implementation review — 2 blocking findings** (posted via `/cockpit:auto`).

This is the cluster-side integration harness (#1024), whose whole purpose is to prove the end-to-end path across the four P1 siblings. As written the harness is a skeleton — every scenario is skipped and the infrastructure is never executed, so the integration job is green but asserts nothing. See the two inline comments.

PR #1027, opened 2026-07-21
feat: #1021 1021-part-cockpit-remote-gates

christrudelpw:
## Summary

Closes #1021

---


packages/orchestrator/src/routes/cockpit-answers.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (!options.writer.isHealthy()) {
+        return reply.status(503).send({
+          error: 'answers-file writer not available',
+          code: 'ANSWERS_FILE_UNAVAILABLE',
+        });
+      }
+
+      try {
+        const parsed = GateAnswerSchema.parse(request.body);
+        if (options.writer.hasDelivered(parsed.deliveryId)) {
```
christrudelpw, line 42: **Non-atomic `deliveryId` dedup (TOCTOU) — violates "appends exactly once".** Two concurrent POSTs carrying the same `deliveryId` (cloud at-least-once redelivery: a retry arrives before the first request responds) both call `writer.hasDelivered(deliveryId)` and get `false`, because neither append has completed yet. Both then `await writer.append()`. The writer serializes appends via its mutex but never re-checks `hasDelivered` inside the critical section — it only calls `dedup.add()` *after* the write — so both lines are appended. The route-level concurrency test only asserts `deliveredCount <= 2` (and notes the second "may be deduped depending on scheduling"), so exactly-once is not actually guaranteed. **Fix:** move the dedup check inside `append()`'s mutex-protected critical section and skip the write if the `deliveryId` is already present.

christrudelpw (commented):
**Implementation review — 1 blocking finding** (posted via `/cockpit:auto`).

The `deliveryId` dedup is not atomic, which violates the "appends exactly once" acceptance criterion for this issue. See the inline comment. Everything else (FIFO retention, N=3 rotation, current-file dedup rebuild, apiKeyStore auth, implicit dispatcher fallback) matches the clarification decisions.

packages/orchestrator/src/routes/cockpit-answers.ts
christrudelpw: Addressed in 0d61507d — please review, and re-open this thread if it still falls short.

PR #1028, opened 2026-07-21
feat: #1023 1023-part-cockpit-remote-gates

christrudelpw:
## Summary

Closes #1023

---


packages/generacy/src/cli/commands/cockpit/doorbell.ts
```diff
@@ -374,6 +386,67 @@ export async function runDoorbell(
...
+  // Answers-file tailer — peer wake source that runs concurrently with
+  // whichever primary source `source-selector` picks. Bound to the same
+  // epicRef as the doorbell; writes gate-answer events to stdout and emits
+  // them into a shared bus so `cockpit_await_events` sees them too. Wrapped
+  // in the same startup-retry envelope as poll-mode acquire so transient
+  // ECONNRESET / rate-limit responses are retried consistently.
+  const acquireForTailer = deps.acquireBus ?? acquireEpicBus;
+  const tailerAcquireOptions: AcquireOptions = {
+    epicRef: form.ref,
+    logger,
```
christrudelpw, line 400: **Cross-epic drops (and rotation/truncation info logs) are silently no-op'd in production — violates clarification Q1 ("logs dropped cross-epic lines at `info` with gateId — not silent-drop").** The `logger` passed here into the answers-file tailer is, in the real CLI path, the default logger defined earlier in this file (~line 338) which declares only `warn` — **no `info` method**. `runDoorbell` is invoked without `deps.logger`, so it falls back to that default. Every tailer call site uses optional chaining `this.logger.info?.(...)` (cross-epic drop in `answers-file-source.ts`, rotation, truncation, waiting-for-dir/file), so in production those resolve to `undefined` and emit nothing anywhere. A cross-epic answer line is therefore dropped **silently**, which is exactly what Q1 says must not happen. The tests only pass because they inject `logger = { warn, info: vi.fn() }`. **Fix:** add an `info` method to the default doorbell logger (writing to stderr, matching `warn`), and add `info` to the `DoorbellDeps.logger` type so the contract is enforced.

christrudelpw (commented):
**Implementation review — 1 blocking finding** (posted via `/cockpit:auto`).

The cross-epic-drop diagnostics required by clarification Q1 are silently no-op'd in production because the doorbell's default logger has no `info` method. See the inline comment. Everything else (wait-for-dir, interleaving replay, warn→logger sink, line-count ceiling) matches the clarification decisions.

packages/generacy/src/cli/commands/cockpit/doorbell.ts
christrudelpw: Addressed in afa78455 — please review, and re-open this thread if it still falls short.

PR #1039, opened 2026-07-23
feat: #1038 1038-issue-1038

### LINKED ISSUE #1038: cockpit gates: read-only gate-status query (MCP tool + route) + stable sweep generation derivation so gateIds coalesce across restarts
Part of the **Cockpit Remote Gates** epic (generacy-ai/generacy-cloud#850) — dogfood follow-up filed from the `--gates=ui` run tracked in generacy-ai/agency#450. Design + wire contracts: [docs/cockpit-remote-gates-plan.md](https://github.com/generacy-ai/tetrad-development/blob/develop/docs/cockpit-remote-gates-plan.md).

Dependency of the agency sweep fix (see the linked agency issue): the `--gates=ui` startup sweep re-drafts and re-opens gates that are already pending because there is no cheap way to ask *"is a gate already open for this issue + type?"* before doing the drafting work, and because sweep-derived `gateId`s don't match live-derived ones.

## Problem 1 — no gate-status query

`cockpit_gate_open` can't serve as an existence check:

- Its input requires the fully-drafted `title`/`body`/`options` (`GateOpenInputSchema`, `packages/generacy/src/cli/commands/cockpit/mcp/gates/schemas.ts:84-107`) — i.e. you must already have run the drafting subagent.
- Its output is `{ gateId, status }` where `status` is only `retained` (relay down) or `open` (emitted), derived from the relay envelope, **not** from cloud gate state (`.../mcp/tools/cockpit_gate_open.ts:107-114`). There is no `answered`/`already-exists` signal.
- It is documented **"Not idempotent"** (`specs/1022-part-cockpit-remote-gates/contracts/cockpit_gate_open.md:106-111`).
- The orchestrator route is fire-and-forget with no lookup (`packages/orchestrator/src/routes/cockpit-gates.ts:86-126`).
- There is **no GET route and no `cockpit_gate_status`/`cockpit_gate_list` tool** — `mcp/server.ts` registers only `cockpit_gate_open` (`:205`) and `cockpit_gate_ack` (`:216`).

**Fix:** add a read-only query — a `GET /cockpit/gates?issueRef=&gateType=` (or by `gateId` prefix) orchestrator route plus a thin `cockpit_gate_status` / `cockpit_gate_list` MCP tool — returning `{ gateId, status: open | answered | absent }` **without requiring the drafted body**. Cloud (Firestore) is the source of truth; decide during plan whether the orchestrator proxies the query over the relay (same path answers ride) or cloud exposes it directly.

## Problem 2 — sweep generation derivation is unstable, so gateIds don't coalesce

Even with a query, the sweep and live paths derive **different** `gateId`s for the same gate, so a cloud upsert can't coalesce them:

- The sweep hard-codes `generation=1` (`packages/claude-plugin-cockpit/commands/auto.md:198` in agency), contradicting the per-`gateType` content-derived discriminator the live path uses (e.g. clarification generation = content hash of the answer set, `auto.md:1356-1358`).
- The durable `generation` discriminator is **not computed today** for `clarification` (no answer-set hash), `artifact-review`/`implementation-review` + `manual-validation` (no head SHA), and `escalation` (no occurrence counter) — the DATA GAPS at `auto.md:1367`. Until derived from durable GitHub state, re-asks across restart/takeover aren't idempotent for those gate types.

`gateId = sha256("<issueRef>:<gateType>:<generation>")[:24]` (`schemas.ts:67-77`), so a stable, durable-GitHub-derived `generation` is required for the same gate to hash identically across sessions. This repo owns `deriveGateKey`/`deriveGateId` and the generation rules — reconcile them so the Problem 1 query returns a match. (The corresponding `generation=1` sweep default is fixed on the agency side.)

## Acceptance criteria

- A read-only gate-status query exists (MCP tool + route) returning `open | answered | absent` for a gate identity **without** the drafted body.
- For `clarification` and `implementation-review` gates, a gate re-derived by the startup sweep produces the **same** `gateId` as the originally-opened gate (generation derived from durable GitHub state, not `1`).
- Contract/schema docs + changeset updated.

christrudelpw:
## Summary

Closes #1038

---


christrudelpw (commented):
**Implementation review (`/cockpit:auto`)** — 1 blocking finding, see the inline comment.

Everything else checks out against the approved clarifications: the 7→3 status collapse (Q2=C), the fail-loud-not-absent unreachable path (Q3=D), the `cockpit_gate_list` non-terminal/project-wide set (Q5=A), the deterministic `{questionNumber, questionText}` generation hash (SC-002), and the flat `z.object(...).strict()` MCP input schemas all match spec.

packages/generacy/src/cli/commands/cockpit/mcp/gates/query-client.ts
```diff
@@ -0,0 +1,197 @@
...
+        `orchestrator returned malformed ${verb} envelope: ${firstLineOr(msg, 'shape mismatch')}`,
+      );
+    }
+  }
+  const bodyText = await readTextSafe(res);
+  const detail = firstLineOr(bodyText, `HTTP ${res.status}`);
+  if (res.status === 400) {
+    throw new QueryInvalidArgsError(detail);
+  }
+  if (res.status >= 500 && res.status < 600) {
```
christrudelpw: **HTTP 500 mis-mapped to `query-unreachable` — the contract requires `internal`.**

The `res.status >= 500 && res.status < 600` branch maps HTTP **500** to `QueryTransportError` → error class `query-unreachable`. But the frozen contract (`specs/1038-issue-1038/contracts/gate-query.md` non-2xx table, and this PR's own `cloud-gate-query-client.ts` header comment) says a route **500 must surface as class `internal`**.

Why it matters: the orchestrator route returns HTTP 500 specifically for a `CloudRequestError` — i.e. the cloud gate endpoint returned a 4xx (unknown clusterId / 422), a non-JSON body, or a 200 with a malformed / shape-drifted envelope. Those are deterministic internal/cluster faults, not transient transport failures. With the current mapping, `decodeResponse` → `QueryTransportError` → `isRetryableGateQueryError` is true → `withRetry` burns 3 attempts (~5 s) on a deterministic error, then `cockpit_gate_status` / `cockpit_gate_list` return `query-unreachable` — which per the `ErrorClass` contract **aborts the `--gates=ui` scope run as if the cloud were down**, masking the real bug (exactly the generacy-cloud-side schema drift this epic can produce).

Fix: map HTTP 500 → class `internal` (non-retryable); keep `502/503/504` → `query-unreachable`. Also update `query-client.test.ts`, which currently pins `[500,502,503,504] → QueryTransportError` and locks the wrong mapping in.

christrudelpw: Addressed in 32d71a1b — please review, and re-open this thread if it still falls short.

PR #1045, opened 2026-07-24
feat: #1043 1043-summary-when-speckit-feature

christrudelpw:
## Summary

Closes #1043

---


packages/orchestrator/src/worker/pr-manager.ts
```diff
@@ -145,6 +146,57 @@ export class PrManager {
...
+            issueNumber: this.issueNumber,
+            currentBranch: branch,
+            canonicalBranch: canonical.branchName,
+            source: canonical.source,
+            anchoringPrNumber: canonical.anchoringPrNumber,
+            action: 'no-op',
+          },
+          'workflow-reentry-branch-mismatch',
+        );
+        return undefined;
```
christrudelpw: **Blocking — no-op stalls PR creation (contradicts Q2=A).** When `resolveIssueBranch` returns a canonical `<N>-*` branch that differs from the current checkout AND has no open PR (`findPRForBranch` → null), this path returns `undefined` and never opens a PR. Scenario: a stale PR-less debris branch `<N>-old` exists on the remote and the worker is on the real work branch (re-entry past create_feature, or the resolver callback unwired — the exact "sole dedup site" case this guard targets) → the resolver picks the debris branch → no-op → the real implementation is left permanently PR-less until manual cleanup. Q2=A says PR-less branches must be *ignored*, not treated as canonical. The dedup test asserting "adoption target has no PR → no-op" cements the stall as intended — reconsider both.

packages/workflow-engine/src/actions/builtin/speckit/lib/feature.ts
```diff
@@ -297,10 +297,54 @@ export async function createFeature(input: CreateFeatureInput): Promise<CreateFe
...
+        err
+      );
+    }
+  }
+
+  // Create branch name using configured pattern (or resolver override / short_name).
+  const derivedBranchName = input.short_name
     ? `${featureNum}-${input.short_name}`
     : buildBranchNameFromPattern(branchConfig, featureNumInt, input.description);
+  const branchName = resolvedBranchName ?? derivedBranchName;
```
christrudelpw, line 335: **Blocking — cold re-entry clobbers the real spec (reintroduces #1043).** Binding `branchName` to the resolver's canonical branch here means the create-path (taken when `featureDir` is absent, i.e. a fresh/cold workspace) will `git checkout -b <canonical> origin/<canonical>` to restore the real branch, then the existing `writeFile(specFile, <template>)` (~L523) overwrites the real spec.md with a stub and pushes it onto the canonical PR — the exact spec-only-regeneration outcome #1043 targets, now landing on the *real* PR instead of a duplicate. The resume path (dir exists) is safe, so only cold-workspace re-entry clobbers, and no test covers it. Guard the create-path so a resolver-supplied canonical branch does not regenerate/overwrite an existing spec.

packages/workflow-engine/src/actions/builtin/speckit/lib/issue-branch-resolver.ts
```diff
@@ -0,0 +1,159 @@
...
+      issueNumber,
+      error: error instanceof Error ? error.message : String(error),
+    });
+  }
+
+  // Step 2: remote branches matching <N>-*, oldest by commit timestamp.
+  const candidateBranches: Array<{ name: string; timestamp: number }> = [];
+  let branchListFailed = false;
+  try {
+    const branches = await github.listBranches(owner, repo);
```
christrudelpw, line 99: **Blocking — un-paginated branch enumeration silently drops branches past page 1.** `listBranches` calls `gh api /repos/.../branches` with no `--paginate`/`per_page`, capping at GitHub's 30-per-page default. Q1=A makes remote branches the *sole* source of truth, so on a repo with >30 branches (generacy has far more), a `<N>-*` branch beyond page 1 is invisible → this resolver returns null when no open PR anchors the issue → createFeature falls back to slug derivation → a fresh divergent `<N>-*` branch + `specs/<N>-*` dir is created, reintroducing the orphaned-slug bug. Silent, no error. Paginate the enumeration (and note the companion `listOpenPullRequests --limit 100` cap on the PR path).

packages/workflow-engine/src/actions/builtin/speckit/lib/issue-branch-resolver.ts
```diff
@@ -0,0 +1,159 @@
...
+    const matching = branches.filter((b) => filter.test(b));
+    for (const name of matching) {
+      let timestamp = Number.POSITIVE_INFINITY;
+      try {
+        const raw = await git.raw([
```
christrudelpw, line 127: **Non-blocking — oldest-branch tiebreak degrades on a cold cache.** Selecting oldest via tip-commit `%ct` on `refs/remotes/origin/<name>` without a prior fetch throws for branches known only via the API (missing remote-tracking refs) → timestamp becomes `+Infinity` → the tiebreak silently falls back to alphabetical (not oldest), defeating the cold-cache resilience Q1=A was justified on. Also tip-commit time ≠ branch age (an actively-updated canonical branch can sort newer than stale debris). Consider fetching first, or ordering by branch creation.

christrudelpw (commented):
## Implementation review — request changes

Automated implementation review for #1043 (deterministic branch/PR dedup). **3 blocking findings** — as written, this fix can *reintroduce* the dup-PR / orphaned-implementation failure it targets (on cold-workspace re-entry and on repos with >30 branches), and can permanently stall PR creation. Details in the inline threads.

Grounded against the operator-approved clarifications: **Q1=A** (remote branches = sole source of truth), **Q2=A** (oldest open PR canonical; PR-less `<N>-*` branches ignored), **Q4=A** (first-derived slug wins), **Q5=A** (dedup applied unconditionally).

PR #1048, opened 2026-07-26
feat: #1047 1047-problem-orchestrator-s-pr

christrudelpw:
## Summary

Closes #1047

---


christrudelpw (commented):
Review of generacy-ai/generacy#1048 — 7 finding(s), 4 blocking.

7 anchored inline; 0 could not be anchored to a diff hunk.

packages/orchestrator/src/worker/pr-feedback-ack-parser.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function parseAcknowledgedFindings(
+  commentBodies: readonly string[],
+): AcknowledgedFindings {
+  const matching = commentBodies.filter(body =>
+    body.includes(BODY_FINDINGS_UNADDRESSED_MARKER),
+  );
+  if (matching.length === 0) return new Set();
+
+  const newest = matching[matching.length - 1]!;
```
christrudelpw, line 34: **BLOCKING** Ack set is always empty in production: `listPrCommentBodies` returns one array element per LINE, not per comment, so the marker line is isolated from the enumeration rows and FR-008/Q6=B is entirely non-functional.

**Failure scenario:** `GhCliGitHubClient.listPrCommentBodies` (gh-cli.ts:795) runs `gh pr view --json comments --jq '.comments[].body'` then `stdout.split('\n')`. gh emits string values raw, so ONE multi-line comment becomes N array elements (verified: `gh pr view 1048 --json body --jq .body` emits 7 lines for one body). The Disposition-C comment is multi-line, so `matching` is the single element `<!-- generacy-cockpit:body-findings-unaddressed -->`, `newest` is that lone line, and `parseEntries` finds zero `- `x` review #N finding M` rows. Verified by direct execution: whole-body input gives Set{'bot:700:1','bot:700:2'}; real line-split input gives Set{} while the marker `.some(includes)` check still returns true. Consequence: after Disposition C the operator removes `blocked:body-finding-unaddressed`, the next cycle re-computes an empty ack set, the same findings gate again, and Disposition C re-fires — the exact permanent label-loop Q6=B was chosen to prevent. All three tests (ack-parser unit + SC-007 integration) mask this by feeding whole comment bodies as single array elements, which the real client never produces.

packages/orchestrator/src/worker/pr-feedback-handler.ts
```diff
@@ -329,6 +387,55 @@ export class PrFeedbackHandler {
       // 7b. Happy path — CLI succeeded AND we have a real commit.
       const shortSha = (await this.getHeadShortSha(checkoutPath)) ?? '<unknown>';
 
+      // #1047 T013: per-finding body-gate. Compute the just-pushed touched
+      // file set, parse any prior Disposition-C acknowledgment marker, and
+      // evaluate `evaluateBodyGate`. On unsatisfied: apply Disposition C
+      // (blocked:body-finding-unaddressed + marker comment) and skip the
+      // reply/resolve loop.
+      const commitTouchedFiles = await this.getCommitTouchedFiles(
```
christrudelpw: **BLOCKING** The gate's touched-file set is the whole PR diff (`origin/<base>..HEAD`), not the fix cycle's commits, so any body finding naming a file already in the PR diff auto-satisfies without the fixer touching it.

**Failure scenario:** research.md D7 and tasks.md T012 specify `getCommitTouchedFiles(checkoutPath, baseSha, headSha)` where baseSha is the pre-fix (cycle-start) SHA; this call site passes `origin/${pr.base.ref}` — the base BRANCH tip. `git diff --name-only origin/develop..HEAD` on this very PR returns all 26 files the PR ever changed. Concrete failure reproducing the #1047 evidence: /cockpit:auto posts a body finding with `**Files:** packages/claude-plugin-cockpit/commands/auto.md` (auto.md IS in the PR diff — this is agency#460 round-1/round-2's auto.md:28 finding). The fixer ignores the body finding and only edits an inline-thread file. `origin/develop..HEAD` still lists auto.md from the implement-phase commits, so the finding is marked satisfied, the gate passes, and Disposition A fires with the cycle falsely complete. FR-003/SC-003/SC-006 are defeated for every body finding naming an in-diff file. Second-order: `git diff A..B` in diff mode is a two-endpoint tree diff, so files changed on the base branch since divergence also appear as touched, adding further false satisfactions. Fix: capture HEAD SHA before spawnClaudeForFeedback/commitAndPushChanges and diff <preFixSha>..HEAD.

packages/orchestrator/src/worker/pr-feedback-handler.ts
```diff
@@ -234,6 +250,35 @@ export class PrFeedbackHandler {
         throw new Error(`Failed to fetch review threads for PR #${prNumber}: ${String(error)}`);
       }
 
+      // 3b. #1047: also fetch review submissions so top-level review-body
+      // findings that name files NOT in the diff still reach the fixer prompt
+      // (inline threads only surface findings anchored to a file+line). Failure
+      // is log-and-continue — the thread path is independent and must still run.
+      try {
+        const reviews = await github.listReviews(owner, repo, prNumber);
+        relevantReviews = reviews.filter(
```
christrudelpw: **BLOCKING** Review bodies bypass the `isTrustedCommentAuthor` filter that guards the inline path, letting any GitHub user inject prompt content and permanently block the fixer loop via Disposition C.

**Failure scenario:** Step 3 filters every inline comment through `isTrustedCommentAuthor(c, 'pr-feedback', ...)` and drops untrusted authors before the CLI sees them (#842/#869/#878). Step 3b filters reviews only by `state` and non-empty body — no trust check, no botLogin/authorAssociation gate. `GET /pulls/{n}/reviews` returns submissions from ANY GitHub user. Attack: while one legitimate trusted inline thread is open (needed to get past Case A/B), an unaffiliated user submits a COMMENTED review whose body contains `<!-- generacy-cockpit:unanchored-findings -->` + `### Finding 1` + `**Files:** does-not-exist.md`. evaluateBodyGate can never satisfy it, so every cycle lands on Disposition C, applies `blocked:body-finding-unaddressed`, skips reply/resolve, and the monitor's `blocked:*` skip gate halts the PR-feedback loop indefinitely. The same untrusted body text is also concatenated into promptInputs and fed to the fixer (only wrapUntrustedData-fenced, not dropped as the inline path requires).

packages/orchestrator/src/worker/pr-feedback-handler.ts
```diff
@@ -764,6 +921,93 @@ Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>`;
...
+        BLOCKED_BODY_FINDING_UNADDRESSED_LABEL,
+      ]);
+    } catch (error) {
+      this.logger.warn(
+        { error: String(error), issueNumber, label: BLOCKED_BODY_FINDING_UNADDRESSED_LABEL },
+        'Failed to add blocked:body-finding-unaddressed label — non-fatal, comment carries the same info',
+      );
+    }
+
+    if (existingCommentBodies.some(b => b.includes(BODY_FINDINGS_UNADDRESSED_MARKER))) {
```
christrudelpw: **BLOCKING** `applyDispositionC` skips posting whenever ANY prior marker comment exists, so the second and later Disposition C never record their findings — contradicting its own contract and freezing operator triage at cycle 1.

**Failure scenario:** contracts/body-findings-unaddressed-marker.md § Idempotency requires skipping only when a marker exists AND the enumeration set is identical, and states 'Multi-cycle marker accumulation is expected. Each Disposition C posts its own marker comment.' The implementation drops the set-equality clause. Scenario: cycle 1 blocks on review #700 finding 1 and posts the marker. Operator hand-fixes and removes the label. Reviewer submits a NEW review #800 with finding 1 naming X.md; the fixer misses it. Gate fails, addLabels fires, `existingCommentBodies.some(includes(MARKER))` is true (the cycle-1 comment), so it returns before postPrComment. No comment ever enumerates bot:800:1, so (a) the operator sees a blocked label whose only marker comment lists stale review-#700 findings, and (b) even with finding #1 fixed, the ack set can never contain bot:800:1, so the gate re-blocks on every subsequent cycle with no escape. The unaddressed set is already computed and passed in — the equality check is available at zero cost.

packages/orchestrator/src/worker/pr-feedback-handler.ts
```diff
@@ -329,6 +387,55 @@ export class PrFeedbackHandler {
...
+      const acknowledged = parseAcknowledgedFindings(existingCommentBodies);
+      const gateResult = evaluateBodyGate({
+        parsedReviews,
+        commitTouchedFiles,
+        acknowledged,
+      });
+
+      if (!gateResult.satisfied) {
+        this.logger.warn(
+          {
```
christrudelpw: Disposition C returns before the reply/resolve loop, contradicting FR-007's stated trigger ('thread resolves succeeded but ...') and invalidating the Q6 premise-correction the whole ack design rests on.

**Failure scenario:** FR-007 and the Q6 rationale both specify the hold fires AFTER the thread resolves succeed (the Q6 answer's load-bearing correction is that removing the label does NOT re-enqueue precisely because totalUnresolvedThreads === 0 makes the monitor hit Case C at pr-feedback-monitor-service.ts:264-285 before the blocked:* check at :328-341). The gate is placed at :395, before step 8's reply-then-resolve loop, and returns. Concrete effect: a cycle where the fixer correctly fixed all inline threads but missed a body finding leaves every inline thread unresolved and unreplied — the reviewer sees no 'Addressed in <sha>' acknowledgment, and the next cycle re-feeds the same already-fixed inline comments to the fixer (redundant token spend, risk of double-editing). It also flips the resume path from the documented 'nothing re-enters until a new review' to immediate re-enqueue on label removal, which is only survivable if the ack set works.

packages/orchestrator/src/worker/pr-feedback-handler.ts
```diff
@@ -234,6 +250,35 @@ export class PrFeedbackHandler {
...
+            totalReviews: reviews.length,
+            relevantReviews: relevantReviews.length,
+          },
+          'Fetched PR reviews for body-consumption path (#1047)',
+        );
+      } catch (error) {
+        this.logger.warn(
+          { error: String(error), prNumber, owner, repo },
+          'Failed to fetch PR reviews (#1047 body-consumption path) — continuing with thread-only inputs',
+        );
```
christrudelpw, line 344: A review carrying only a body finding and no inline threads never reaches the fixer at all — SC-001's literal acceptance criterion fails for the pure case.

**Failure scenario:** PrFeedbackMonitorService builds unresolvedThreadIds from threads.filter(t => !t.isResolved) (:203) and only enqueues when that set is non-empty; the handler additionally returns at Case A when unresolvedThreadCount === 0 (:277) and at Case B when unresolvedComments.length === 0, both before reviewBodyItems is constructed. SC-001's literal acceptance criterion — 'post a review with gh pr review --request-changes --body "<finding about file X>" where X is not in the diff, assert the next commit touches X' — therefore fails whenever that review carries no inline comments: nothing enqueues and the body is never fetched. The fix works only for mixed reviews that also contain at least one trusted inline thread. Either widen the monitor trigger or state the scope limit explicitly in the spec and changeset.

packages/workflow-engine/src/actions/github/client/gh-cli.ts
```diff
@@ -679,6 +681,49 @@ export class GhCliGitHubClient implements GitHubClient {
...
+    const allowedStates: ReadonlySet<ReviewSubmissionState> = new Set([
+      'APPROVED',
+      'CHANGES_REQUESTED',
+      'COMMENTED',
+      'DISMISSED',
+      'PENDING',
+    ]);
+
+    return data.map(r => {
+      if (!allowedStates.has(r.state as ReviewSubmissionState)) {
```
christrudelpw, line 721: `listReviews` throws on any unrecognized `state` value, discarding ALL reviews on that PR and silently disabling the entire #1047 path.

**Failure scenario:** The allowedStates guard runs inside data.map(...), so a single review whose state is outside the hard-coded five (a future GitHub state, or any casing/shape drift) throws for the whole call. The handler's step-3b catch converts that into parsedReviews = [] / relevantReviews = [] with a warn log, so review bodies stop reaching the prompt AND the FR-003 gate silently stops gating — the original #1047 bug returns with no error surfaced to the operator. Since the consumer only cares about CHANGES_REQUESTED/COMMENTED, an unknown state should be passed through or that single review skipped, not made fatal for the batch.

packages/orchestrator/src/worker/pr-feedback-ack-parser.ts
christrudelpw: Addressed in ce006b6b — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/pr-feedback-handler.ts
christrudelpw: Addressed in ce006b6b — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/pr-feedback-handler.ts
christrudelpw: Addressed in ce006b6b — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/pr-feedback-handler.ts
christrudelpw: Addressed in ce006b6b — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/pr-feedback-handler.ts
christrudelpw: Addressed in ce006b6b — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/pr-feedback-handler.ts
christrudelpw: Addressed in ce006b6b — please review, and re-open this thread if it still falls short.

packages/workflow-engine/src/actions/github/client/gh-cli.ts
christrudelpw: Addressed in ce006b6b — please review, and re-open this thread if it still falls short.

PR #1050, opened 2026-07-26
feat: #1049 1049-problem-pr-review-feedback

christrudelpw:
## Summary

Closes #1049

---

### DIFF
--- a/packages/orchestrator/src/routes/__tests__/pr-webhooks.test.ts
+++ b/packages/orchestrator/src/routes/__tests__/pr-webhooks.test.ts
@@ -50,15 +50,21 @@ function createMockMonitorService(
 /**
  * Create a minimal valid GitHub PR review webhook payload.
  *
- * #1049: accepts an optional `merged` field on `pull_request` (default `false`)
- * so tests can exercise the merged-PR gate.
+ * #1049: GitHub sends `SimplePullRequest` on `pull_request_review` /
+ * `pull_request_review_comment` events, which omits `merged` but carries
+ * `merged_at` (null on unmerged PRs). Tests accept an optional `merged_at`
+ * to exercise the merged-PR gate against the real wire shape. The optional
+ * `merged` field is retained for a small set of tests that document the
+ * synthetic fallback path.
  */
 function createWebhookPayload(
   overrides: Partial<GitHubPrReviewWebhookPayload> & {
     merged?: boolean;
+    merged_at?: string | null;
+    state?: string;
   } = {},
 ): GitHubPrReviewWebhookPayload {
-  const { merged, ...rest } = overrides;
+  const { merged, merged_at, state, ...rest } = overrides;
   const base: GitHubPrReviewWebhookPayload = {
     action: 'submitted',
     review: {
@@ -73,8 +79,9 @@ function createWebhookPayload(
       body: 'Fixes #100',
       head: { ref: '100-add-tests', sha: 'abc123' },
       base: { ref: 'main' },
-      state: 'open',
+      state: state ?? 'open',
       ...(merged !== undefined ? { merged } : {}),
+      ...(merged_at !== undefined ? { merged_at } : {}),
     },
     repository: {
       owner: { login: 'test-org' },
@@ -1204,6 +1211,14 @@ describe('PR Webhook Route - POST /webhooks/github/pr-review', () => {
 
   // ==========================================================================
   // #1049: prMerged propagation from webhook payload → PrReviewEvent
+  //
+  // GitHub sends `SimplePullRequest` on pull_request_review and
+  // pull_request_review_comment events. Verified against
+  // @octokit/webhooks-types@7.6.1 and @octokit/openapi-webhooks-types@11.0.0:
+  // `merged` is NOT a field on SimplePullRequest, but `merged_at` and `state`
+  // both ARE. So the production wire shape is (a) merged_at !== null on a
+  // merged PR, (b) merged_at === null on an unmerged PR, and (c) no `merged`
+  // key at all. The realistic-payload cases below drive that shape directly.
   // ==========================================================================
 
   describe('#1049 merged-PR flag propagation', () => {
@@ -1211,8 +1226,11 @@ describe('PR Webhook Route - POST /webhooks/github/pr-review', () => {
       server = await buildServer({ monitorService });
     });
 
-    it('SC-006 webhook half: pull_request.merged=true → PrReviewEvent.prMerged=true', async () => {
-      const payload = createWebhookPayload({ merged: true });
+    it('SC-006 webhook half: realistic merged PR (merged_at non-null, state=closed, no `merged` key) → prMerged=true', async () => {
+      const payload = createWebhookPayload({
+        merged_at: '2026-07-26T12:34:56Z',
+        state: 'closed',
+      });
       const rawBody = JSON.stringify(payload);
 
       await server.inject({
@@ -1230,8 +1248,11 @@ describe('PR Webhook Route - POST /webhooks/github/pr-review', () => {
       );
     });
 
-    it('pull_request.merged=false → PrReviewEvent.prMerged=false', async () => {
-      const payload = createWebhookPayload({ merged: false });
+    it('realistic unmerged PR (merged_at=null, state=open, no `merged` key) → prMerged=false', async () => {
+      const payload = createWebhookPayload({
+        merged_at: null,
+        state: 'open',
+      });
       const rawBody = JSON.stringify(payload);
 
       await server.inject({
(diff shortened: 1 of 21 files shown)


packages/orchestrator/src/routes/pr-webhooks.ts
```diff
@@ -113,6 +113,8 @@ export async function setupPrWebhookRoutes(
         prBody: payload.pull_request.body || '',
         branchName: payload.pull_request.head.ref,
         source: 'webhook',
+        // #1049 (FR-008): boundary-sanitize for backward-compat with test doubles.
+        prMerged: payload.pull_request.merged ?? false,
```
christrudelpw: **BLOCKING** The merged-PR gate reads `payload.pull_request.merged`, a field GitHub NEVER sends on `pull_request_review` / `pull_request_review_comment` webhooks, so `prMerged` is always false and gate G1 (FR-008, Q2=B) never fires in production.

**Failure scenario:** GitHub delivers the reduced `SimplePullRequest` object on review events, not the full `PullRequest`. Verified against two independent schema sources in this workspace: `@octokit/webhooks-types@7.6.1` types `PullRequestReviewSubmittedEvent.pull_request` as `SimplePullRequest`, and diffing `SimplePullRequest` against `PullRequest` shows `merged` is one of exactly 12 dropped fields (additions, changed_files, comments, commits, deletions, maintainer_can_modify, mergeable, mergeable_state, merged, merged_by, rebaseable, review_comments); `merged_at` and `state` ARE present. `@octokit/openapi-webhooks-types@11.0.0`'s `webhook-pull-request-review-submitted.pull_request` likewise carries `merged_at` and `state` and no `merged`. Because the new type declares `merged?: boolean` (types/monitor.ts:139) and the read is `?? false`, this is silent — no type error, no runtime error.

Concrete failure: a reviewer posts an inline follow-up comment on a MERGED PR whose issue is at `completed:validate` → webhook arrives with no `merged` key → `prMerged=false` → G1 skipped → the widened FR-001 guard now ACCEPTS the issue (`completed:*` is evidence) where the old `agent:*`-only guard rejected it → assignee + trust checks pass → enqueue → `pr-feedback-handler.ts:127-138` checks out `pr.head.ref` on the reused bootstrapped checkout, where `repo-checkout.ts:110` fetches WITHOUT `--prune`, resets to the pre-merge tip, and pushes at `:670` — RESURRECTING the branch `cockpit merge` deleted.

That is exactly generacy-cloud#883, which occurred 69 seconds after its parent PR merged. This PR ships the widening that makes the defect reachable (spec FR-008 says so explicitly) together with a gate that cannot fire. Root cause is research.md D5, which asserts 'The webhook payload always carries `merged: boolean` … applicable to pull_request_review.submitted and pull_request_review_comment.created' — that claim is false.

Fix: `payload.pull_request.merged ?? (payload.pull_request.merged_at != null)`. `merged_at` is present on the wire and is already carried in the type at monitor.ts:140 as 'not consumed'.

packages/orchestrator/src/services/smee-receiver.ts
```diff
@@ -413,6 +413,8 @@ export class SmeeWebhookReceiver {
       prBody: pr.body ?? '',
       branchName: pr.head.ref,
       source: 'webhook',
+      // #1049 (FR-008): boundary-sanitize for backward-compat.
+      prMerged: pr.merged ?? false,
```
christrudelpw: **BLOCKING** Same dead read as pr-webhooks.ts:117 — `pr.merged ?? false` on a smee-delivered `pull_request_review` payload is always false, so the merged-PR gate is inert on the path this cluster actually uses.

**Failure scenario:** smee relays the verbatim GitHub payload, whose `pull_request` is `SimplePullRequest` and therefore has no `merged` key (see the pr-webhooks.ts:117 finding for the schema evidence). The local type widening at smee-receiver.ts:400 adds `merged?: boolean` to the cast, which makes the code compile but does not make the field exist on the wire. Every review event arriving over the smee→webhook bus — the primary delivery path for this cluster per the doorbell design, confirmed live by `source=smee reason=startup-smee-selected` — yields `prMerged: false`, including reviews on merged PRs. Same resurrected-branch outcome as the webhook path. Fix: `pr.merged ?? (pr.merged_at != null)`, adding `merged_at?: string | null` to the inline cast.

packages/orchestrator/src/routes/__tests__/pr-webhooks.test.ts
```diff
@@ -1192,4 +1201,72 @@ describe('PR Webhook Route - POST /webhooks/github/pr-review', () => {
...
+  // #1049: prMerged propagation from webhook payload → PrReviewEvent
+  // ==========================================================================
+
+  describe('#1049 merged-PR flag propagation', () => {
+    beforeEach(async () => {
+      server = await buildServer({ monitorService });
+    });
+
+    it('SC-006 webhook half: pull_request.merged=true → PrReviewEvent.prMerged=true', async () => {
+      const payload = createWebhookPayload({ merged: true });
```
christrudelpw: **BLOCKING** The SC-006 webhook-half test synthesizes a `pull_request.merged` field that GitHub never sends, so the suite is green while the gate it certifies is dead in production — and a companion test encodes the bug as expected behaviour.

**Failure scenario:** `createWebhookPayload({ merged: true })` (lines 61/77) injects a key absent from the real `pull_request_review` payload. The assertion `expect.objectContaining({ prMerged: true })` therefore proves only that the route copies a field the wire never carries — it cannot fail for the reason it exists. Worse, the companion test at line 1252 ('merged field omitted → prMerged defaults to false') asserts that the REAL production payload shape produces `prMerged: false`, i.e. the suite explicitly pins the broken behaviour as correct. There is no test anywhere that drives the route with a realistic merged-PR review payload (`merged_at` non-null, `state: 'closed'`, NO `merged` key) and asserts the gate fires. Fix the two source findings above, then re-anchor these tests on `merged_at` and add a realistic-payload case.

packages/orchestrator/src/services/pr-feedback-monitor-service.ts
```diff
@@ -421,6 +438,74 @@ export class PrFeedbackMonitorService {
...
+          owner, repo, prNumber, issueNumber, gate, source,
+          probeError: err instanceof Error ? err.message : String(err),
+        },
+        `PR-feedback event dropped by ${gate} gate (probe failed)`,
+      );
+      return;
+    }
+
+    if (unresolvedThreads >= 1) {
+      this.logger.info(
```
christrudelpw, line 517: `dropWithGateLog` adds an unconditional GraphQL `getPRReviewThreads` call and an unthrottled `info` line per poll cycle for every unlinked / non-orchestrated / unassigned open PR — the exact steady-state GraphQL burn and log-spam that clarification Q3's rationale used to keep the wrong-cluster gate at `debug`.

**Failure scenario:** `pollRepo` (:707) iterates every open PR in every monitored repo each cycle (60s default) and feeds it to `processPrReviewEvent`; `deduplicatePrsByIssue` explicitly retains PRs that resolve to no issue. Before this PR a `no-link` / `not-orchestrated` / `assignees-empty` drop cost zero API calls and one `debug` line. Now each such PR costs one GraphQL query per cycle, forever: a long-lived human or dependabot PR in any of the 8 repos in this cluster's MONITORED_REPOS is ~60 GraphQL queries/hour each, charged to the same shared 5k/hr budget this account already exhausts. Additionally, an unlinked human PR carrying one unresolved review thread emits an `info` line every 60 seconds indefinitely.

This is a gap in the approved Q3 answer rather than an implementation error — Q3=B exempted the wrong-cluster gate on exactly this amplification argument, but the same argument applies verbatim to the three lifted gates on the poll path, which the answer did not consider. Note the file already solves this shape for thread counts at :281-299 via transition-edge logging (`lastUnresolvedThreadCount` / `isTransition`); the new gate logs have no equivalent guard.

Mitigation that preserves Q3/Q4 conformance: memoize the drop-gate outcome per `owner/repo#prNumber` so the probe and the `info` line fire on the transition edge rather than every cycle, and/or skip the thread probe entirely when `source === 'poll'`.

christrudelpw (commented):
Review of generacy-ai/generacy#1050 — 4 finding(s), 3 blocking.

4 anchored inline; 0 could not be anchored to a diff hunk.

packages/orchestrator/src/routes/pr-webhooks.ts
christrudelpw: Addressed in 81543840 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/smee-receiver.ts
christrudelpw: Addressed in 81543840 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/routes/__tests__/pr-webhooks.test.ts
christrudelpw: Addressed in 81543840 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/pr-feedback-monitor-service.ts
christrudelpw: Addressed in 81543840 — please review, and re-open this thread if it still falls short.

Issue #883, opened 2026-07-09
orchestrator: PR-feedback loop never terminates — handler replies but never resolves threads, so the unresolved-thread trigger re-fires forever (reply spam doubles per cycle)


christrudelpw:
Found during the cockpit v1 integration smoke test (generacy-ai/tetrad-development#88), finding #37 — discovered on the PR-feedback loop's first successful end-to-end run (post-#878/#879 deploy).

## Observed (christrudelpw/sniplink#4 / PR #14, 2026-07-09 10:52–11:04Z)

The loop's first cycle worked: 5 trusted unresolved threads → handler → Claude CLI (~3.6 min) → 4 files changed → commit `86d5f20` pushed → 5 thread replies ("I've addressed this feedback in the latest commit. Please review the changes.") → label cleared → success.

Then it did it again. And again:

- **Cycle 2** (claimed 10:56:57, ~22 s after cycle 1 completed): same 5 threads re-detected — **replies do not resolve threads**, and the monitor's trigger is `unresolvedThreads > 0`. Full CLI run (~2.2 min) → `No changes to commit — skipping commit/push` → **posted 10 more replies** (`threadCount: 10` — one per *comment*, including cycle 1's own replies) → success.
- **Cycle 3** enqueued 11:00:08 the moment cycle 2's item left in-flight. Operator's PR view at this point: three identical "I've addressed this feedback…" replies stacked on each thread and counting.

Nothing in the system can end this: the trigger state (unresolved threads) is one no code path transitions, so the loop re-fires at poll cadence forever — a full Claude CLI invocation plus a reply batch per ~5 minutes, with the reply batch growing each cycle (reply-per-comment: 5 → 10 → …). The #879 dedupe is working correctly throughout (drops while in-flight, structured `reason: "in-flight"` lines); this is not a dedupe bug — the *work generator* never terminates.

Manual intervention that stopped it (and proves the fix mechanism): resolved all 5 threads via `resolveReviewThread` GraphQL mutations using the cluster's own App credential — all succeeded. Next poll saw 0 unresolved; loop quiesced.

## Root cause

The handler treats "reply posted" as completing the feedback, but the monitor treats "thread unresolved" as feedback pending. Those are different state planes and the handler never writes to the one the monitor reads. A successful cycle must transition its own trigger.

## Proposal

[...]

Issue #437, opened 2026-04-11
[refactor-spawn][wave-5] Add lint rule forbidding direct child_process.spawn


christrudelpw:
## Goal

Add a lint rule forbidding direct `child_process.spawn` outside the two sanctioned `ProcessFactory` implementations and `AgentLauncher` internals. Prevents new direct-spawn call sites from sneaking into the codebase post-refactor.

## Scope

- Add an ESLint rule (or custom check, per repo conventions) that forbids importing or calling `child_process.spawn`, `spawn`, `exec`, `execFile`, and `fork` outside these allow-listed file paths:
  - `packages/orchestrator/src/worker/claude-cli-worker.ts` (`defaultProcessFactory`)
  - `packages/orchestrator/src/conversation/process-factory.ts` (`conversationProcessFactory`)
  - Any new files under an `agent-launcher/` internals directory, if one exists
  - Test files and test fixtures (allow by path glob, not inline disable)
- Allow-list via file-path matching, not inline comments — so new files are forbidden by default.
- Rule runs in CI as part of `pnpm lint`.

## Acceptance criteria

- `pnpm lint` passes on `develop` after this rule lands (assuming Waves 1-4 have already cleaned up direct spawns).
- A deliberately regressing branch (adding `child_process.spawn` to a random file) fails lint with a clear error message pointing at the rule.
- The lint rule itself is documented in the eslint config or in a short comment explaining the rationale.

## Out of scope

- Runtime sandbox enforcement.
- Enforcing `ProcessFactory` or `AgentLauncher` usage via type-level tricks.

## Dependencies

- Depends on Waves 1-4 all landing (otherwise the rule would fail on existing unmigrated code).

## References

- Parent tracking: #423

christrudelpw:
## 📋 Specification Stage

| Phase | Status | Started | Completed |
|-------|--------|---------|-----------|
| specify | ✅ complete | — | — |
| clarify | ✅ complete | 2026-04-12T20:32:42.934Z | 2026-04-12T20:33:38.157Z |
| plan | ⏳ pending | — | — |
| tasks | ⏳ pending | — | — |
| implement | ⏳ pending | — | — |
| validate | ⏳ pending | — | — |

**Status**: ✅ Complete
**Started**: 2026-04-12T20:32:42.934Z
**Completed**: 2026-04-12T20:33:38.157Z
**PR**: https://github.com/generacy-ai/generacy/pull/452

christrudelpw:
[...]

PR #1052, opened 2026-07-27
feat: #1051 1051-problem-after-speckit-pr

christrudelpw:
## Summary

Closes #1051

---


packages/orchestrator/src/worker/push-guard.ts
```diff
@@ -0,0 +1,144 @@
...
+      branch,
+      owner,
+      repo,
+      issueNumber,
+    };
+  }
+
+  // Row 3 / Row 5: branch missing on remote. Both open-PR and no-PR cases
+  // fall through here.
+  if (!branchPresent) {
```
christrudelpw: **BLOCKING** Row 3 refuses the FIRST PUSH OF EVERY NEW ISSUE. `pr === null` + branch-not-yet-on-origin falls into `if (!branchPresent)` and returns `refuse: 'branch-missing'` — the exact case Q2 said must be ALLOW, and the case the comment 8 lines below claims is allowed.

**Failure scenario:** Brand-new issue #2000. `createFeature` only runs `git.checkoutLocalBranch(branchName)` — it never pushes (zero push call sites in speckit/lib/feature.ts), so `2000-foo` does not exist on origin. `context.branch` is set before phase-loop is entered, so the loop-entry guard at phase-loop.ts:228 runs BEFORE the first phase. Inside evaluatePushGuard: `findPRForBranchAnyState` -> null (no PR yet); `defaultRemoteBranchExists('2000-foo')` -> `git ls-remote --heads origin 2000-foo` exits 0 with EMPTY stdout, so push-guard.ts:65 `return stdout.trim() !== ''` -> false. Verified live: `git ls-remote --heads origin definitely-does-not-exist-xyz` -> exit=0, stdout empty. No throw, so the fail-open catch at :93 never engages. `pr === null` skips rows 1 and 2; `!branchPresent` fires row 3 -> REFUSE. phase-loop.ts:237-240 then calls handlePhaseLoopPushRefused and returns `{ completed: false, gateHit: false }`, which claude-cli-worker maps to 'Workflow stopped due to phase failure'. NOT ONE PHASE EXECUTES. Every new issue queued to any cluster dies at specify with `agent:error`.

This contradicts spec.md:63 ('`null` (no PR in any state) -> allow (first-push case, no false positive)') and the code's own comment at push-guard.ts:141-142. CI is green because push-guard.test.ts:147 encodes the WRONG expectation ('no PR + branch MISSING -> refuse') and push-guard.test.ts:141 mislabels 'no PR + branch PRESENT' as 'the first-push case' — which it is not; on a genuine first push the branch is absent. No phase-loop or pr-manager test covers this; the pre-existing phase-loop tests pass only because their github mocks lack `findPRForBranchAnyState`, so the call throws a TypeError inside the try and fail-open returns allow — the mocks are accidentally masking the bug.

Fix: row 3 must distinguish `pr === null` (no PR ever -> ALLOW, first push) from `pr && pr.state === 'open'` + branch missing (-> refuse 'branch-missing'). Then correct push-guard.test.ts:141/147 — as written those two tests pin the defect as intended behaviour.

packages/orchestrator/src/worker/pr-manager.ts
```diff
@@ -110,6 +117,28 @@ export class PrManager {
...
+          repo: this.repo,
+          issueNumber: this.issueNumber,
+          branch,
+          github: this.github,
+          git: {
+            remoteBranchExists: (b) => defaultRemoteBranchExists(b, this.checkoutPath),
+          },
+        });
+        if (decision.kind === 'refuse') {
+          await this.handlePushRefused(decision);
```
christrudelpw, line 177: **BLOCKING** The refusal returns `false` from commitAndPush, but the caller `commitPushAndEnsurePr` then runs `ensureDraftPr()` UNCONDITIONALLY — so the guard blocks the push and the very next line opens the duplicate `Closes #N` PR anyway. The PR's headline symptom is not prevented.

**Failure scenario:** pr-manager.ts:59-63 is:

    async commitPushAndEnsurePr(phase, options?) {
      const hasChanges = await this.commitAndPush(phase, options?.message);
      const prUrl = await this.ensureDraftPr();      // <-- unconditional
      return { prUrl, hasChanges };
    }

The guard at :127 fires, handlePushRefused runs, and :139 `return false` exits commitAndPush — correct so far. Control returns to :61 and `ensureDraftPr()` executes regardless. On a fresh PrManager (`prNumber`/`prUrl` undefined — a new worker run) :248 does not early-return; `tryAdoptCanonicalPr` uses `findPRForBranch` (OPEN-ONLY, :196) which returns null for a merged PR; the second `findPRForBranch` at :273 returns null too; :287 `createPullRequest` opens a NEW draft PR titled `feat: #900 900-...` with body `Closes #900` — against a merged/closed branch.

That is precisely generacy-cloud#883 and #886, the two field instances this PR exists to stop, and a direct violation of FR-003 ('the job MUST exit without opening a new PR'). Note the guard makes it *worse* than status quo in one respect: the push is skipped, so the duplicate PR now points at a branch the worker did not even update.

Fix: `commitPushAndEnsurePr` must short-circuit — propagate the refusal out of `commitAndPush` (a discriminated result, not a bare `false`) and skip `ensureDraftPr()` on refusal.

packages/orchestrator/src/worker/pr-manager.ts
```diff
@@ -110,6 +117,28 @@ export class PrManager {
...
+          return false;
```
christrudelpw: **BLOCKING** `return false` on refusal is indistinguishable from the legitimate 'nothing to commit' return at :110, so the phase loop cannot see the refusal: it marks the phase COMPLETE, re-refuses on every remaining phase, exits `completed: true`, and flips the PR ready-for-review — with zero commits having reached origin.

**Failure scenario:** phase-loop.ts:668 is `const { prUrl, hasChanges } = await prManager.commitPushAndEnsurePr(phase);` and uses the result only to set `context.prUrl`. There is no refusal channel in the return type and no abort. So after a refusal the loop calls onPhaseComplete (adding `completed:<phase>`), advances to the next phase, and calls commitPushAndEnsurePr again — which refuses again, emitting ANOTHER `{event:'push-refused'}` warn and ANOTHER removeLabels/addLabels round-trip.

Three consequences:
(1) SC-002's 'exactly one warn line per refusal' becomes N lines for one logical condition — the assertion holds only in the unit test, never in the field.
(2) The loop reaches :926 with `completed: true`, so claude-cli-worker runs `onWorkflowComplete()` and `markReadyForReview()`. The issue is labelled a completed workflow and the PR leaves draft, despite nothing being pushed.
(3) `agent:error` is added by handlePushRefused and then left in place while the workflow simultaneously reports success — contradictory terminal state.

Compare the loop-ENTRY guard at phase-loop.ts:238-239, which does abort correctly (`return { completed: false, ... }`). The pre-push site needs equivalent treatment; the asymmetry is the bug.

packages/orchestrator/src/worker/push-guard.ts
```diff
@@ -0,0 +1,144 @@
...
+  const { owner, repo, branch, issueNumber, github, git } = input;
+
+  let pr: Awaited<ReturnType<GitHubClient['findPRForBranchAnyState']>>;
+  let branchPresent: boolean;
+  try {
+    [pr, branchPresent] = await Promise.all([
+      github.findPRForBranchAnyState(owner, repo, branch),
+      git.remoteBranchExists(branch),
+    ]);
+  } catch {
```
christrudelpw, line 113: **BLOCKING** The documented fail-open catch is unreachable for the PR lookup: `findPRForBranchAnyState` returns `null` on non-zero exit instead of throwing, so a rate-limited `gh` call is silently reclassified as 'no PR' and the guard ALLOWS the exact resurrection push it exists to block.

**Failure scenario:** gh-cli.ts:936-938 is `if (result.exitCode !== 0) { return null; }`. So when `gh pr list -R o/r --head 900-foo --state all` fails — 403 rate-limit, network blip, auth expiry — the method returns null rather than throwing. The try/catch at push-guard.ts:88-98 is never entered.

Concrete: PR #905 for issue #900 is MERGED and its branch still exists on origin (merged without --delete-branch, or already resurrected by an earlier run), so `branchPresent === true`. The `gh pr list` call is rate-limited — realistic here, since the cluster and the operator share one account's 5k/hr budget and GraphQL exhausts first. Result: `pr === null` skips rows 1 and 2, `branchPresent === true` skips row 3, and :143 returns `allow`. The worker pushes onto the merged branch and (per the pr-manager finding above) opens the duplicate PR.

The guard's own comment at :74-76 justifies fail-open with 'FR-001 and FR-004 are the correctness gates'. Neither holds here: FR-001's `fetch --prune` only bites when the remote branch is GONE (it is present in this scenario), and FR-004 is being cut as invalid (see the retraction on the issue). In the branch-retained case the guard is the ONLY gate — and it opens.

Fix: make the lookup failure distinguishable from 'no PR'. Either have the guard call a throwing variant, or return a tri-state (`{found: PR} | {none} | {error}`) and treat `error` explicitly. Silent null-on-error is the wrong contract for a safety gate.

packages/orchestrator/src/__tests__/repo-checkout-cross-issue.test.ts
```diff
@@ -0,0 +1,149 @@
+/**
```
christrudelpw: **BLOCKING** Delete this file and cut FR-004 entirely — it regression-tests a bug that never existed. The 'cross-issue contamination' premise is retracted; the commit it was inferred from is a MERGE commit, not a single-parent commit.

**Failure scenario:** The issue body claimed `d8e392ca` was 'a single-parent commit, not a merge — the worker had #880's files in its working tree and committed them onto #879's branch'. That is false. `gh api repos/generacy-ai/generacy-cloud/commits/d8e392ca --jq '.parents[].sha'` returns TWO parents: `c3cbe0e4` (#879's pre-merge tip) and `5542e900` (the squash-merge of PR #881, i.e. how #880's work reached develop). It is a routine base merge of `develop` into the branch — the #864 pre-phase hook that `repo-checkout.ts:126-145` exists to serve.

The `added` statuses that produced the misreading are an API artifact: GitHub diffs a merge commit against parent 1 only, so files arriving via parent 2 render as `added`. Confirmed structurally — `specs/` at `c3cbe0e4` contains `879-problem-any-4xx-from` and no `880-*` dir, exactly as a base merge predicts. Corroborated by the second field instance: generacy-cloud#886 resurrected `884-problem-refreshaccesstoken` with no new commit and NO foreign files at all; if checkout reuse leaked working trees, it would have shown there too.

Full retraction: generacy-ai/generacy#1051 comment 5094709961. Cut this file, FR-004, SC-003, and issue acceptance-criterion 4. Note plan.md already chose a NULL implementation for FR-004 ('reuse the existing `git reset --hard HEAD` + `git clean -fd` — no new subsystem'), so deleting it removes a test for a mechanism that was never broken, not any real protection. tasks.md isolates this cleanly as Phase 5 with no interdependencies — the cut is mechanical.

Also delete research.md R4 rather than amending it: its independent rationale ('the reset resolved to the pre-merge tip, which by coincidence had issue-B's files') is self-refuting — files already present on the tip cannot render as `added`.

CLAUDE.md
```diff
@@ -559,3 +559,25 @@ See [/workspaces/tetrad-development/docs/DEVELOPMENT_STACK.md](/workspaces/tetra
...
 - Changeset: `.changeset/1047-pr-feedback-review-bodies.md` — `@generacy-ai/workflow-engine` **minor** (new public `listReviews` method + new `blocked:body-finding-unaddressed` label vocabulary) + `@generacy-ai/orchestrator` **patch** (internal handler behavior change; no new exports). Single file, both bumps.
+
+## Worker branch-resurrection + cross-issue contamination fix (#1051, planning phase)
+
+- Bug: after a speckit PR merges with `--delete-branch`, an orchestrator worker re-entering the same bootstrapped checkout resurrects the deleted remote branch, commits files scoped to a **different** issue onto it, and opens a duplicate PR claiming `Closes #<already-closed>`. Field instance: `generacy-ai/generacy-cloud#883` — PR #882 (`Closes #879`) squash-merged with branch deletion at 12:52:23; commit `d8e392ca` (`chore(speckit): complete validate phase for #879`) authored 63 s later on the resurrected `879-problem-any-4xx-from`; PR #883 opened 6 s after that. `git compare develop...d8e392ca` → `ahead_by=9, behind_by=1` — nine orphan commits. Every file in the orphan commit belongs to **#880**, not #879.
+- Root causes (compose to produce the outcome):
```
christrudelpw, line 566: **BLOCKING** This appended planning note writes the retracted contamination claim into permanent repo documentation, where it will mislead every future reader and agent long after the PR is forgotten.

**Failure scenario:** The added text states as established fact: 'commits files scoped to a **different** issue onto it' and 'Every file in the orphan commit belongs to **#880**, not #879.' Both are wrong — see the retraction finding above and generacy#1051 comment 5094709961. CLAUDE.md is loaded into context for every agent session in this repo, so unlike a stale spec file this actively propagates the error into future work: the next agent to touch repo-checkout.ts will read that checkout reuse leaks working trees between issues and design around a hazard that does not exist.

The third root-cause bullet is wrong for the same reason and must go with it — 'the reset restores the pre-merge tip which by coincidence had another issue's files staged (via createFeature's per-issue spec-dir cd)' describes a mechanism that was never observed.

Fix: rewrite the note to describe ONE defect — branch resurrection producing a duplicate PR — and drop every contamination sentence plus the FR-004 bullet. Keep the FR-001/002/003/005 material, which is accurate and genuinely useful. Also correct the timing detail while editing: the note says PR #883 opened '6 s after' the commit; the two field instances measured merge-to-duplicate-PR at 69 s and 71 s, which is the number worth recording.

packages/workflow-engine/src/actions/github/client/gh-cli.ts
```diff
@@ -923,6 +923,46 @@ export class GhCliGitHubClient implements GitHubClient {
...
   }
 
+  async findPRForBranchAnyState(owner: string, repo: string, branch: string): Promise<PullRequest | null> {
+    const result = await this.executeGh([
+      'pr', 'list',
+      '-R', `${owner}/${repo}`,
+      '--head', branch,
+      '--state', 'all',
+      '--json', 'number,title,body,state,isDraft,headRefName,baseRefName,labels,createdAt,updatedAt',
+      '--limit', '1',
```
christrudelpw: `--limit 1` returns the NEWEST PR on the branch regardless of state, so a merged PR hidden behind a newer closed one reports `pr-closed` — defeating the merged-precedence the decision matrix is explicitly built for.

**Failure scenario:** Live fixture, exactly the shape this PR is about: `gh pr list -R generacy-ai/generacy-cloud --head 884-problem-refreshaccesstoken --state all --limit 1` returns only `[{number:886, state:"CLOSED", createdAt:"2026-07-27T16:53:34Z"}]`. The MERGED #885 (createdAt 2026-07-26) is dropped by the limit. So push-guard.ts:102 `pr.state === 'merged'` is skipped and :115 `pr.state === 'closed'` fires, yielding reason `pr-closed`.

Not a correctness break — both rows return `kind: 'refuse'` and every consumer branches only on `decision.kind`, with `reason` used for logging only. The push is still blocked. But it defeats the stated intent at push-guard.ts:100-101 ('Short-circuits row 3 so a merged PR whose branch was deleted still produces the more diagnostic pr-merged reason'), and an operator debugging a resurrection sees the least informative of the two reasons.

Note also that contracts/find-pr-for-branch-any-state.md justifies `--limit 1` with 'picking the newest merged PR over a stale older one is correct' — factually wrong on this fixture, where the newest is CLOSED and the merged one is older. Fix: raise the limit and prefer a merged PR when several exist on the branch, or state plainly in the contract that `reason` is best-effort.

packages/orchestrator/src/services/label-monitor-service.ts
```diff
@@ -332,6 +332,32 @@ export class LabelMonitorService {
...
+    // — a `resume` event on already-closed generacy-cloud#879 sailed through
+    // this method and reached the git layer, producing the observed
+    // resurrection + duplicate-PR outcome. Zero mutations on drop
+    // (no enqueue, no markProcessed, no label mutation).
+    //
+    // Fallback: if `fetchedIssue` is null (the try/catch above swallowed a
+    // fetch error), the gate does NOT fire — better to enqueue a possibly-
+    // closed issue than to drop a definitely-open one on transient `gh`
+    // failure. FR-002 pre-push guard catches the closed case downstream.
+    if (fetchedIssue && fetchedIssue.state === 'closed') {
```
christrudelpw, line 346: The gate fires at ENQUEUE time, not at worker-claim time, so it would not have caught the observed repro — SC-004's '100% of dispatched phase jobs whose target issue is closed' overstates what ships.

**Failure scenario:** `processLabelEvent` is the enqueue step. The actual dispatch is `worker-dispatcher.ts:318` `const item = await this.queue.claim(workerId)` -> :382 'Dispatching worker for claimed item', with no issue-state read anywhere between claim and worker spawn (no getIssue call in that file). So: issue open at T+0 -> resume enqueued, gate passes -> PR squash-merged and issue auto-closed at T+30s -> worker claims the still-queued item at T+40s -> full checkout + phase + commit on a closed issue.

The field evidence lands squarely in that window: PR #882 merged and #879 closed at 12:52:23; the commit was authored at 12:53:26, 63 s later. FR-005 as implemented would most likely NOT have stopped it — FR-002's guard is the real catch. research.md R8 does acknowledge this composition, so the implementation matches the design; the defect is that SC-004's wording claims dispatch-time coverage the gate does not provide. Either re-word SC-004 to say 'enqueue-time', or add the state check at claim time where 'dispatch' actually happens.

The in-scope gate logic itself is correct and was verified by mutation: reverting only label-monitor-service.ts to develop makes the two closed-state tests fail, including the resume case, and `getIssue` normalizes state via `.toLowerCase()` at gh-cli.ts:154 so `=== 'closed'` matches.

packages/orchestrator/src/services/label-monitor-service.ts
```diff
@@ -332,6 +332,32 @@ export class LabelMonitorService {
...
+      this.logger.info(
```
christrudelpw, line 347: Four other enqueue paths bypass this gate entirely, including a base-advance re-validate whose fingerprint matches the observed repro more closely than the path that was gated.

**Failure scenario:** Paths that never reach processLabelEvent, none with a closed-issue check: (1) `server.ts:745-762` base-advance resume — builds `{command:'continue', queueReason:'resume', description:'Base advanced to <sha>; re-validating'}` and enqueues directly; its source `base-advance-monitor-service.ts:216-260` selects open PRs carrying `failed:validate` and uses `const issueNumber = pr.number`. That resume flavour produces exactly the worker commit shape seen in the field ('complete validate phase for #N'). (2) `worker-dispatcher.ts:267` lease-expiry re-enqueue. (3) `worker-dispatcher.ts:447` post-complete rearm. (4) `pr-feedback-monitor-service.ts:406` — #1049 gates merged PRs there, not closed issues with an open PR.

`clarification-answer-monitor-service.ts:236` and `merge-conflict-monitor-service.ts:182` are incidentally safe because their discovery calls `listIssuesWithLabel`, which passes `--state open` (gh-cli.ts:198).

Not blocking — FR-005 names only processLabelEvent, so this is scope, not a defect. But it must be recorded, because the changeset and CLAUDE.md note currently read as 'closed issues can no longer be dispatched', which is not what ships. Worth a follow-up issue naming path (1) specifically.

christrudelpw (commented):
Review of generacy-ai/generacy#1052 — 11 finding(s), 6 blocking.

11 anchored inline; 0 could not be anchored to a diff hunk.

packages/orchestrator/src/services/__tests__/label-monitor-service.closed-issue.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    const { svc, logger, phaseTracker, queueManager } = makeSvc(factory);
+
+    const result = await svc.processLabelEvent(makeEvent('resume'));
+
+    expect(result).toBe(false);
+
+    const dropCall = findInfoCall(logger, 'dropped', 'issue-closed');
+    expect(dropCall).toBeDefined();
+    const payload = dropCall![0] as Record<string, unknown>;
+    expect(payload).toMatchObject({
```
christrudelpw, line 184: The resume-drop test omits `phase` from its log assertion, so a regression that dropped the field for resume events only would stay green.

**Failure scenario:** The contract requires `{dropped, issueNumber, eventType, phase, owner, repo}` on every drop line, and the process test asserts `phase: 'speckit-feature'` at :155 — but the resume test's `toMatchObject` at :179-185 lists only dropped/issueNumber/eventType/owner/repo. A plausible future refactor ('a resume has no phase, it has a gate') that emitted `phase: undefined` for resumes would keep the suite green while operators lost the field they grep on. Add `phase` to the resume assertion.

Also worth stating in the PR description: of the 5 new tests in this file only 2 discriminate against develop (process+closed, resume+closed); the three open/fallback cases pass identically pre- and post-fix. That is expected for over-fire guardrails, but the raw test count overstates coverage.

specs/1051-problem-after-speckit-pr/research.md
```diff
@@ -0,0 +1,178 @@
...
+
+**Alternatives considered**:
+
+- *`git checkout -B ... origin/<branch>` unconditionally* — invasive rewrite of the switch logic; would still not detect a merged-and-deleted branch without a prior fetch --prune (the tracking ref would still be present pre-prune).
+- *Test remote branch existence explicitly with `git ls-remote --heads origin <branch>` inside `switchBranch`* — one extra network call per phase entry vs. `--prune`'s piggyback on the existing fetch. `--prune` is strictly cheaper.
```
christrudelpw, line 14: R1 supports `--prune` by quoting the git manpage text for `--no-tags`, not `--prune`.

**Failure scenario:** The quoted passage is 'By default, tags that point at objects that are downloaded from the remote repository are fetched and stored locally. This option disables this automatic tag following.' That documents `--no-tags`. It says nothing about pruning stale remote-tracking refs and does not support the decision it is cited for. The conclusion (add `--prune` at both sites) is correct and independently justified by the composition argument in the same section — only the citation is wrong. Replace it with the actual `--prune` text ('Before fetching, remove any remote-tracking references that no longer exist on the remote') or drop the quote.

packages/orchestrator/src/worker/pr-manager.ts
```diff
@@ -46,13 +65,31 @@ export class PrManager {
...
+  async commitPushAndEnsurePr(phase: WorkflowPhase, options?: { message?: string }): Promise<CommitResult> {
+    const outcome = await this.commitAndPush(phase, options?.message);
+    if (outcome.kind === 'refused') {
+      // Do NOT call ensureDraftPr — otherwise the guard blocks the push but
+      // the very next line opens a duplicate PR that claims `Closes #N`
+      // against a merged/deleted branch (PR #1052 review Finding 2).
+      const { reason, prNumber, branch, owner, repo, issueNumber } = outcome.refusal;
+      return {
+        hasChanges: false,
+        pushRefused: { reason, prNumber, branch, owner, repo, issueNumber },
```
christrudelpw, line 89: **BLOCKING** CI Build fails: TS2322 + TS2345. The new `'pr-lookup-failed'` reason was added to the guard's union (Finding 4 fix) but never propagated into pr-manager's local refusal type, so the file does not compile.

**Failure scenario:** Real CI failure on e26a2ee1, job `integration`, step `Build (packages)`, exit code 2:

  src/worker/pr-manager.ts(88,24): error TS2322: Type '"pr-merged" | "pr-closed" | "branch-missing" | "pr-lookup-failed"' is not assignable to type '"pr-merged" | "pr-closed" | "branch-missing"'.
    Type '"pr-lookup-failed"' is not assignable to type '"pr-merged" | "pr-closed" | "branch-missing"'.
  src/worker/pr-manager.ts(176,40): error TS2345: Argument of type '{ kind: "refuse"; reason: "pr-merged" | "pr-closed" | "branch-missing" | "pr-lookup-failed"; ... }' is not assignable to parameter of type '{ reason: "pr-merged" | "pr-closed" | "branch-missing"; ... }'.

This is the tail of the timed-out round: `push-guard.ts` gained the fourth reason value, and `handlePushRefused` / the `pushRefused` field on `CommitResult` still declare the original three.

Fix: widen both declaration sites to the guard's exported reason type rather than re-listing the literals. Import the union from push-guard (e.g. `Extract<PushGuardDecision, { kind: 'refuse' }>['reason']`) so a fifth reason can never desynchronize these types again — re-typing the literals by hand is exactly what produced this failure.

Also confirm the FR-003b label branch and the `event: 'push-refused'` log handle `pr-lookup-failed` deliberately: it is the one reason that means "we could not determine safety", not "we determined it is unsafe", and an operator needs to tell those apart.

packages/orchestrator/src/worker/push-guard.ts
```diff
@@ -0,0 +1,184 @@
+import { execFile } from 'node:child_process';
```
christrudelpw: **BLOCKING** CI Lint fails: `node:child_process` is banned repo-wide by `no-restricted-imports`, and this new file is not on the allowlist. One error, and it fails the whole `ci` job.

**Failure scenario:** Real CI failure on e26a2ee1, job `ci`, step `Lint (packages)`:

  packages/orchestrator/src/worker/push-guard.ts
    1:1  error  'node:child_process' import is restricted from being used. Direct child_process usage is forbidden. Use ProcessFactory or AgentLauncher instead. See #437  no-restricted-imports

The rule is declared in `.eslintrc.json` under `rules.no-restricted-imports` (paths: `child_process`, `node:child_process`) and switched `off` for an explicit allowlist in `overrides[0].files`, which contains `claude-cli-worker.ts`, `process-factory.ts`, `cli-utils.ts`, `relay-bridge.ts`, `repo-checkout.ts`, `base-merge.ts`, `merge-conflict-handler.ts` and others. That is why `repo-checkout.ts` uses `execFile` freely and this new file cannot. Tests are also exempt via `overrides[1]`, which is why `push-guard.test.ts` is clean and the failure appears only for the source file.

PREFERRED FIX — do not extend the allowlist. `evaluatePushGuard` already takes the git call as an injected dependency (`git: { remoteBranchExists }` on `PushGuardInput`), so the guard itself needs no process access at all; only the `defaultRemoteBranchExists` helper does. Move that helper into `repo-checkout.ts`, which is already allowlisted, already owns every other `git` invocation in the worker, and is already the module all three call sites import for checkout work. `push-guard.ts` then holds pure decision logic with zero Node process imports, and the three call sites pass `repoCheckout.remoteBranchExists` instead of the local default.

FALLBACK: add `packages/orchestrator/src/worker/push-guard.ts` to `overrides[0].files`. This is one line and unblocks CI, but it widens an architectural rule for a file that does not need the exemption — please prefer the move.

Note for the changeset/PR description: the `ci` job runs `Lint (packages)` but the engine's own validate phase does not, so lint breakage of this kind cannot be caught before CI on this workflow.

christrudelpw (commented):
Review of generacy-ai/generacy#1052 — 2 finding(s), 2 blocking.

2 anchored inline; 0 could not be anchored to a diff hunk.

packages/orchestrator/src/worker/pr-manager.ts
christrudelpw: Addressed in 571618b4 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/push-guard.ts
christrudelpw: Addressed in 571618b4 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/push-guard.ts
```diff
@@ -0,0 +1,162 @@
...
+ * one refusal log per refusal).
+ */
+export async function evaluatePushGuard(input: PushGuardInput): Promise<PushGuardDecision> {
+  const { owner, repo, branch, issueNumber, github, git } = input;
+
+  // PR #1052 review Finding 4: split fail-isolation per lookup so a `gh`
+  // failure does not silently reclassify as "no PR". Kick both lookups off
+  // in parallel and await individually so one failure does not cancel the
+  // other's result.
+  const prPromise = github.findPRForBranchAnyState(owner, repo, branch);
```
christrudelpw: **BLOCKING** Both lookup invocations sit OUTSIDE the try blocks, so a SYNCHRONOUS throw at the call site escapes the guard entirely and crashes the whole phase loop. 84 tests fail on this; lint and build now pass, so this is the only thing keeping CI red.

**Failure scenario:** Round 2 split the combined `try { await Promise.all([...]) }` into per-lookup try blocks to fix Finding 4. But the split left the CALLS unprotected:

    const prPromise = github.findPRForBranchAnyState(owner, repo, branch);   // <-- line 84, no try
    const branchPromise = git.remoteBranchExists(branch);                    // <-- line 85, no try
    let pr;
    try { pr = await prPromise; } catch { ... }                              // only the AWAIT is guarded

A method that throws synchronously never produces a promise, so the `try` around the `await` is never entered and the error propagates out of `evaluatePushGuard` altogether.

Observed on CI for 571618b4 — `Test (packages)` and `Test (integration)`, 84 failures:

    TypeError: github.findPRForBranchAnyState is not a function
      ❯ evaluatePushGuard        src/worker/push-guard.ts:84:28
      ❯ PhaseLoop.executeLoopInner  src/worker/phase-loop.ts:230:40
      ❯ PhaseLoop.executeLoop       src/worker/phase-loop.ts:145:25
      ❯ ClaudeCliWorker.handle      src/worker/claude-cli-worker.ts:657:42

Whole suites are taken out: `claude-cli-worker.test.ts` (specify→clarify gate, bugfix gates, continue-after-gate, CLI-crash handling, validate pass/fail, all five markReadyForReview cases) and every `pr-feedback-handler-body-flow.test.ts` case from #1047.

THIS IS NOT ONLY TEST DEBT. The contract at :70-75 promises fail-isolation, and this hole voids it for the entire class of synchronous throws — any `GitHubClient` implementation that lacks the new method, or that validates arguments and throws before returning a promise, now takes down the phase loop at entry instead of failing open. The mocks are the messenger, not the defect.

Note this is the same shape as the finding raised on generacy-cloud#885 (an IIFE whose `finally` could run before the outer assignment when `fetch()` throws synchronously) — a protected region that guards the settle but not the invocation.

Fix (both lines):

    const prPromise = Promise.resolve().then(() => github.findPRForBranchAnyState(owner, repo, branch));
    const branchPromise = Promise.resolve().then(() => git.remoteBranchExists(branch));

or equivalently wrap each invocation in its own try. Keep the per-lookup catch split from Finding 4 — that part is right; only the invocation needs to move inside.

Separately, add `findPRForBranchAnyState` to the shared GitHubClient test double so the mocks exercise the real path rather than relying on it being absent. Do that AFTER the fix above, not instead of it — patching only the mocks turns a red suite green while leaving the fail-open hole in production.

christrudelpw (commented):
Review of generacy-ai/generacy#1052 — 1 finding(s), 1 blocking.

packages/orchestrator/src/worker/push-guard.ts
christrudelpw: Addressed in 08096980 — please review, and re-open this thread if it still falls short.

PR #1055, opened 2026-07-27
feat: #1053 1053-problem-gateid-pure-function

christrudelpw:
## Summary

Closes #1053 (US1 + US2 only; primary end-to-end fix requires the agency-side companion — see below).

## What ships

- `deriveGateKey(issueRef, gateType, generation)` in `@generacy-ai/cockpit` gains an optional fourth argument `runId?: string`. When passed, appended as `:${runId}` to the pre-image so a re-run of the same natural gate produces a fresh `gateId` and stops colliding with a terminal cloud doc from a prior run. When omitted, the derivation is byte-for-byte compatible with pre-#1053.
- `cockpit_gate_open` / `cockpit_gate_ack` schemas accept optional `runId`. `cockpit_gate_open` threads the explicit `runId` through the derivation and logs `runIdSource: 'explicit' | 'unset'` at `info` (`event: 'cockpit_gate_open.runid-source'`).
- `askedAt` hoisted above the retry boundary via a per-`gateId` in-memory cache so within-run retries produce byte-identical wire frames — US2 correctness no longer depends on cloud `gateId`-keyed dedup.

## Post-review changes

- **Removed the process-scoped `INSTANCE_NONCE` fallback** on the missing-`runId` path. Shipping it as the de-facto default reintroduced the same cross-run collision this PR exists to close (two `/cockpit:auto` runs in one MCP-server process would share the nonce) and desynced against the cloud's positional-3-tuple `generationFromGateKey` read-side parser. On the default path, the tool now derives the pre-#1053 3-tuple `gateKey` — a no-op that leaves the schema surface ready but the per-run discriminator inactive until the sibling agency PR threads `runId`.
- `cockpit-gate-open-runid.test.ts` scenario 3 (`askedAt` hoist) rewritten to drive a fake clock (`vi.useFakeTimers` + `vi.setSystemTime`) that advances 5 s between calls, so the assertion discriminates the hoist rather than passing on a same-millisecond coincidence.

## Known limitation — explicit `runId` path is not end-to-end today

The default path (no `runId`) derives the pre-#1053 3-tuple `gateKey` and is fully cloud-compatible on both read and write. **The explicit-`runId` write path (4-segment `gateKey`) is not yet symmetrical with the cloud read side.** Specifically:

- `cockpit_gate_status({issueRef, gateType, generation})` derives the lookup key from the 3-tuple positionally — a `gateId` written with a 4-segment `gateKey` will not be found and the tool returns `absent`.
- `cockpit_gate_list` decodes `generation` from `gateKey` via `generationFromGateKey`, which greedy-slices after the 2nd colon — a listed row for a 4-segment `gateKey` reports `generation: "<generation>:<runId>"` instead of `<generation>`.

The schema surface is shipped now so the agency-side `/cockpit:auto` companion (FR-006) and the cloud read-side companion (`cockpit-gate-query.ts` derivation, tracked separately) can land against a stable input contract. Until BOTH companions merge, callers should NOT pass explicit `runId` — the shipping default (no `runId`) is the supported path.

## Deferred / out of scope

- **FR-004 / FR-005 / FR-007** — terminal-collision detection, the new `'terminal-collision'` `ErrorClass` value, and ack-path parity. Gated on generacy-cloud#887. Follow-up PR under `workflow:speckit-bugfix`. Q5 → A's answer flagged that adding the union value will break exhaustive switches in #1015, #1020, #1024; deferring this pushes that friction to a later, more contended landing, but keeps this PR small and independently landable. The `'terminal-collision'` `ErrorClass` value is **not** added to `mcp/errors.ts` in this PR — an unused enum value is dead code today; it lands with the sibling that emits it.
- **FR-006** — `/cockpit:auto` skill-side handler that passes the auto-run id through as `runId`. Ships in the sibling `agency/packages/claude-plugin-cockpit/commands/auto.md` PR. **Until that lands, the default path in this PR is a no-op** (`runIdSource: 'unset'`).
- **Cloud read-side 4-segment support** — `cockpit-gate-query.ts` `deriveGateKey` / `generationFromGateKey` positional parsing needs to become runId-aware so `cockpit_gate_status` and `cockpit_gate_list` correlate 4-segment writes. Tracked in generacy-cloud (companion issue TBD). Without this, the explicit-`runId` write path lands but is not observable end-to-end (see "Known limitation" above).

---


packages/generacy/src/cli/commands/cockpit/mcp/tools/cockpit_gate_open.ts
```diff
@@ -48,9 +67,30 @@ export function cockpitGateOpen(
     }
 
     const s = parsed.data;
-    const gateKey = deriveGateKey(s.issueRef, s.gateType, s.generation);
+    // #1053 — Fold the per-run discriminator into the `gateKey` pre-image.
+    // Explicit runId wins; missing runId falls back to the per-process
+    // INSTANCE_NONCE so non-auto callers get within-caller stability without
+    // silently regressing to the pre-#1053 colliding behaviour.
+    const effectiveRunId = s.runId ?? INSTANCE_NONCE;
```
christrudelpw: **BLOCKING** No caller anywhere passes `runId`, so 100% of real gate opens take the INSTANCE_NONCE fallback — which is per-MCP-PROCESS, not per-run. That is precisely option B, which Q1 rejected. Two auto runs in one Claude session still collide, so #1053 is not actually fixed on the shipped path.

**Failure scenario:** Line 74 is `const effectiveRunId = s.runId ?? INSTANCE_NONCE;` and `INSTANCE_NONCE` is a module-level `crypto.randomBytes(8).toString('hex')` (mcp/event-bus.ts:72). The MCP server runs over StdioServerTransport — one process per Claude Code session, NOT one per auto run. And nothing supplies runId: `grep -rn runId /workspaces/agency/packages/claude-plugin-cockpit/commands/auto.md` returns ZERO matches, and tasks.md T013 records 'Files edited by this issue: NONE in this repo' for the playbook, deferring FR-006 to the agency companion.

So the default path is the only path, and it has per-process scope:

COLLISION STILL REPRODUCES — operator runs `/cockpit:auto --gates=ui christrudelpw/snappoll#1`; the phase-queue P2 gate opens, is answered, and acks to `applied`. In the SAME session they re-run `/cockpit:auto` on the same epic. `runId` is still undefined, INSTANCE_NONCE is unchanged, so gateKey and gateId are byte-identical — and generacy-cloud message-handler.ts:905 emits `[relay] Ignoring gate-open for terminal gate ... (status: applied)`. That is the exact failure this PR says it closes.

CONVERSE FAILURE — an MCP-server restart mid-run flips the nonce, minting a second gateId for the same natural gate. The original row stays open in the inbox forever, and an operator answer on it arrives carrying a gateId absent from the loop's `openGates` map, so D.12 routes it nowhere. That is the US2 AC-3 violation Q1 rejected option B for, now reachable by default.

The PR body says 'Closes #1053', but on the shipped default the collision is only avoided when the two runs happen in different MCP-server processes — which is not something the operator controls or can observe.

Fix: either land the agency-side runId threading in lockstep (so the explicit path is the real path), or make the fallback run-scoped rather than process-scoped. Shipping a per-process fallback as the de facto default reintroduces the rejected option under a different name.

packages/generacy/src/cli/commands/cockpit/mcp/tools/cockpit_gate_open.ts
```diff
@@ -48,9 +67,30 @@ export function cockpitGateOpen(
...
+    const runIdSource: 'explicit' | 'fallback-instance-nonce' =
+      s.runId !== undefined ? 'explicit' : 'fallback-instance-nonce';
+    const gateKey = deriveGateKey(s.issueRef, s.gateType, s.generation, effectiveRunId);
```
christrudelpw: **BLOCKING** The write side now derives a 4-segment gateKey but the cloud's READ side still derives from the 3-tuple, so every `cockpit_gate_status` lookup misses and returns 'absent' — breaking the pre-draft dedup check the auto loop depends on.

**Failure scenario:** Verified asymmetry. WRITE: line 77 `deriveGateKey(s.issueRef, s.gateType, s.generation, effectiveRunId)` -> doc id = sha256(4-segment)[:24]. READ: generacy-cloud `services/api/src/services/cockpit-gate-query.ts:85` declares `export function deriveGateKey(issueRef: string, gateType: string, generation: string)` — THREE parameters — and :143 does `const gateKey = deriveGateKey(issueRef, gateType, generation); const gateId = deriveGateId(gateKey);` then `getCockpitGateRef(db, orgId, gateId).get()`. The cluster's `cockpit_gate_status` ships only `{issueRef, gateType, generation}` and the orchestrator forwards it verbatim.

The doc-id lookup therefore misses, `if (!snap.exists) return { gateId: null, status: null }` fires, and even on a hash coincidence the re-verify `if (doc.gateKey !== gateKey)` also returns null.

Concrete: /cockpit:auto opens a clarification gate for an issue and the operator has not answered. On the next wake, auto.md dispatch step 0 calls `cockpit_gate_status({issueRef, gateType, generation})`. Pre-change that returned `{status:'open'}` and the plugin skipped re-drafting. Post-change it returns 'absent', so the drafting subagent re-runs on EVERY wake and the `openGates` reuse record — which auto.md:285 makes mandatory for D.12 routing — is never created. The `status:'answered'` FR-009 escape hatch can never fire either.

auto.md:283 states the invariant this breaks verbatim: 'the pre-draft cockpit_gate_status({issueRef, gateType, generation}) check names the same three inputs, so sweep-derived and live-derived gateIds coalesce.'

plan.md:226 dismissed the read side ('Q2 -> A chose the synchronous rejection mechanism ... so no read-side extension is needed') — that reasoning only covered FR-004 detection and missed the pre-existing dedup query. Q3=A's 'no cloud change' premise holds for VALIDATION (gateKey is `z.string().min(1)`) but not for the cloud's derivation and parsing, which are positional.

christrudelpw (commented):
Review of generacy-ai/generacy#1055 — 5 finding(s), 3 blocking.

5 anchored inline; 0 could not be anchored to a diff hunk.

packages/generacy/src/cli/commands/cockpit/mcp/tools/cockpit_gate_open.ts
christrudelpw: **BLOCKING** The appended runId corrupts the `generation` value returned by `cockpit_gate_list`, because the cloud recovers generation by greedily slicing gateKey after the SECOND colon.

**Failure scenario:** generacy-cloud `cockpit-gate-query.ts:107-111`:

    export function generationFromGateKey(gateKey: string): string {
      const firstColon = gateKey.indexOf(':');
      const secondColon = gateKey.indexOf(':', firstColon + 1);
      return gateKey.slice(secondColon + 1);
    }

The greedy slice is deliberate — generation itself may contain ':' (e.g. `spec-review:<sha>`) — and list mode maps every row with `generation: generationFromGateKey(doc.gateKey)`. The orchestrator passes it through untouched.

With gateKey now `christrudelpw/snappoll#1:phase-queue:P2:christrudelpw-snappoll-1-20260727-200458`, `cockpit_gate_list` returns generation = `P2:christrudelpw-snappoll-1-20260727-200458` instead of `P2`.

Any consumer comparing a listed gate's generation against the live-derived discriminator now mismatches on every row. That includes auto.md's UI startup sweep, which per auto.md:285 has only `{gateId, gateType, generation, status}` from the list to reconstruct a gate record — so the sweep cannot recognise its own in-flight gates after a restart.

This is the same root cause as the finding above: appending to a positionally-parsed string is not a private change. Together they mean Q3=A ('fold into gateKey, no cloud change') is not implementable as specified without also updating the cloud's derivation and parsing — which is worth raising explicitly, since the whole point of Q3=A was to avoid cross-repo coupling. Either coordinate the cloud change, or PREPEND/encode the runId in a position the existing parsers ignore.

packages/generacy/src/cli/commands/cockpit/mcp/__tests__/cockpit-gate-open-runid.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      expect(meta.issueRef).toBe(ISSUE_REF);
+
+      // (c) runId itself is NOT in the log metadata (data-model E-3 privacy).
+      expect(Object.keys(meta)).not.toContain('runId');
+    } finally {
+      infoSpy.mockRestore();
+    }
+  });
+
+  it('scenario 3: two calls with the same runId + input produce byte-identical askedAt', async () => {
```
christrudelpw: The two tests guarding the askedAt hoist and the cross-run round-trip do not discriminate — both pass against develop.

**Failure scenario:** Scenario 3 (:134) asserts `expect(firstAskedAt).toBe(secondAskedAt)` after two awaited calls with a synchronous mock fetch. On develop the code is `askedAt: s.askedAt ?? new Date().toISOString()` at millisecond resolution, and two back-to-back in-process calls routinely land in the same millisecond — so it passes unfixed. It is a coin-flip, not a regression guard.

The orchestrator test at `packages/orchestrator/src/__tests__/cockpit-gates-integration.integration.test.ts:294` never invokes the MCP tool at all: it hand-builds both frames (`const gateKeyA = deriveGateKey(issueRef, gateType, generation, runIdA)` at :304) and POSTs them to a route that relays verbatim, so `expect(gateIdA).not.toBe(gateIdB)` asserts its own fixture arithmetic. The 'byte-identical frames' case POSTs the same literal `bodyA` twice with an explicit `askedAt`. Reverting T006/T007 in cockpit_gate_open.ts — the actual fix — leaves both green.

Only scenarios 1 and 2 genuinely discriminate. Make scenario 3 drive the real tool with a controlled clock and assert the cache is consulted.

packages/generacy/src/cli/commands/cockpit/mcp/tools/cockpit_gate_open.ts
```diff
@@ -14,7 +14,9 @@
  * Wire: tetrad-development/docs/cockpit-remote-gates-plan.md § "Wire contracts".
  * Error mapping: contracts/error-mapping.md (mirror of `gates/client.ts`).
  */
+import { getLogger } from '../../../../utils/logger.js';
 import { wrapToolBoundary, type ToolResult } from '../errors.js';
+import { INSTANCE_NONCE } from '../event-bus.js';
```
christrudelpw: Q5=A (new ErrorClass 'terminal-collision') is not implemented — the union is unchanged.

**Failure scenario:** `mcp/errors.ts:22` still lists only 'invalid-args' | 'wrong-kind' | 'unknown-gate' | 'not-an-epic' | 'gate-refusal' | 'transport' | 'query-unreachable' | 'invalid-cursor' | 'not-worker' | 'contended' | 'claim-conflict' | 'scope-not-found' | 'internal'. tasks.md defers it under 'Notes on Deferred Work' along with FR-004/005/007, gated on generacy-cloud#887.

That is a defensible reading of Q2's decoupling — an error class nothing can emit is dead code — and I am not asking for it to be forced in. Recording it because the approved answer said A and the deferral should be a conscious trade, not a silent omission. Note the clarification answer flagged that adding the value 'will break exhaustive switches in #1015, #1020 and #1024'; deferring pushes that friction to a later, more contended landing rather than taking it now while the union has one new consumer. Either is fine — just make it explicit in the PR description.

packages/generacy/src/cli/commands/cockpit/mcp/gates/schemas.ts
```diff
@@ -102,6 +108,14 @@ export const GateOpenInputSchema = z
...
     /** ISO-8601; defaulted to now() by the tool when omitted. */
     askedAt: z.string().datetime().optional(),
+    /**
+     * Optional per-run discriminator (#1053). When passed, folded into `gateKey`
+     * as a trailing `:${runId}` segment so re-runs against the same natural
+     * gate produce a fresh `gateId`. When omitted, the tool mints a per-process
+     * fallback from `INSTANCE_NONCE` — never validated as `default(...)` here so
+     * the source of the fallback is observable at the tool boundary.
+     */
+    runId: z.string().min(1).optional(),
```
christrudelpw, line 131: **BLOCKING** The JSDoc directly above `runId` still documents the REJECTED INSTANCE_NONCE design as the specified contract, while the implementation now does the opposite. This is the first thing the FR-006 implementer will read, and believing it silently preserves the collision #1053 exists to close.

**Failure scenario:** The block comment above `runId: z.string().min(1).optional()` states: 'When omitted, the tool mints a per-process fallback from `INSTANCE_NONCE` — never validated as `default(...)` here so the source of the fallback is observable at the tool boundary.'

That is no longer true and was explicitly rejected on review. The implementation at `schemas.ts:78` now returns the plain 3-tuple base when runId is undefined, and this PR's own changeset records INSTANCE_NONCE as 'rejected on review'. The fix commit updated the tool and its tests but not the field-level contract.

This is the canonical doc surfaced at the MCP boundary. An implementer wiring `/cockpit:auto` (spec FR-006) reads it, concludes omitting `runId` already yields a per-process discriminator, and reasonably omits it for gates they consider process-local — preserving the exact cross-run collision, with no test catching it because scenario 2 asserts the 3-tuple IS correct behaviour.

Fix: rewrite the comment to state that omitting `runId` yields the legacy 3-tuple key with NO run discrimination, and that callers MUST pass it to get the #1053 fix.

Otherwise this PR is approved to merge as a deliberate no-op — the askedAt hoist (verified working, cache consulted above the retry boundary) and the INSTANCE_NONCE removal are both correct and independently worth landing. The remaining read-side work is now tracked in #1059.

christrudelpw (commented):
Review of generacy-ai/generacy#1055 — 1 finding(s), 1 blocking.

PR #1056, opened 2026-07-27
feat: #1054 1054-problem-when-worker-dies

christrudelpw:
## Summary

Closes #1054

---


packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -51,6 +63,54 @@ redis.call('SET', KEYS[3], '1', 'EX', tonumber(ARGV[1]))
...
+if tonumber(ARGV[2]) < tonumber(ARGV[3]) then
+  return 3
+end
+
+redis.call('HDEL', KEYS[1], ARGV[1])
+if redis.call('HLEN', KEYS[1]) == 0 then
+  redis.call('DEL', KEYS[1])
+end
+
+redis.call('SREM', KEYS[3], ARGV[1])
```
christrudelpw: **BLOCKING** RECLAIM_ORPHAN_SCRIPT does `SREM in-flight` AND `ZADD pending` together, breaking the `in-flight = pending ∪ claimed` invariant and re-opening the double-enqueue hole for every reclaimed item.

**Failure scenario:** The invariant is explicit in the existing code. `ENQUEUE_IF_ABSENT_SCRIPT` does SADD in-flight + ZADD pending. `CLAIM_SCRIPT` moves pending→claimed and LEAVES the in-flight member. `release()`'s retry branch (:596-601) does `hdel/del/zadd` with NO srem — and only `complete()` and the dead-letter branch SREM, because those are the two paths where the item actually leaves the pipeline.

The new script does both:

    redis.call('SREM', KEYS[3], ARGV[1])
    redis.call('ZADD', KEYS[4], tonumber(ARGV[4]), ARGV[5])

so the item lands in pending while ABSENT from `orchestrator:queue:in-flight-items`.

Concrete, using this incident's own evidence that the monitors re-enqueue every ~5 min:
  t0        reaper reclaims -> pending holds member M1 (queueReason 'resume', attemptCount 1, priority 0.<ts>); in-flight empty
  t0+5min   monitor calls enqueueIfAbsent -> SISMEMBER returns 0 -> SADD + ZADD member M2 (attemptCount 0, different queueReason/priority/enqueuedAt)

Redis ZSETs key on the MEMBER STRING, and M1 != M2, so these are two distinct pending entries for one issue. `pollOnce` claims one per replica with no per-issue guard, so two replicas — or one replica across two cycles after `complete()` clears the single in-flight member — run the same issue twice: duplicate branch, duplicate PR, duplicate phase work.

That is the #879 double-enqueue class this invariant exists to prevent, and it would be caused by the fix for a wedge, which is a bad trade.

Fix: delete the SREM. A reclaimed item is legitimately still in flight once re-enqueued — exactly what `release()`'s retry branch already models.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.orphan-reclaim.test.ts
```diff
@@ -0,0 +1,485 @@
...
+ *
+ * Reuses the stateful mock pattern from
+ * `redis-queue-adapter.enqueueIfAbsent.test.ts:16-125`, extended with
+ * `EXISTS`/`SCAN`/`HGETALL`/`HLEN`/`DEL`/`SREM`/`ZADD` behaviour and a
+ * `defineCommand('reclaimOrphan', ...)` Lua stub simulating the return-code
+ * contract in `contracts/reclaim-orphan-script.md`.
+ */
+
+interface MockState {
+  pending: Map<string, { score: number; member: string }>;      // itemKey → ZSET entry
```
christrudelpw: **BLOCKING** The 485-line regression suite never executes the Lua script — `defineCommand` is a no-op and `reclaimOrphan` is a hand-written JS re-implementation — and the mock's pending map is keyed by itemKey, which structurally hides the duplicate-member bug above.

**Failure scenario:** `interface MockState { pending: Map<string, {score, member}>; // itemKey -> ZSET entry }` with `zadd: state.pending.set(parsed.itemKey, {...})` (:34-38) gives pending UPSERT-BY-ITEMKEY semantics. Real Redis ZADD keys on the member string. So the duplicate-member hazard from the SREM finding shows as ONE entry in the mock and TWO in production — the mock cannot express the bug, and the suite's own assertion plus its comment ('either a new enqueue succeeds OR the current pending entry blocks it (correct...)') encodes the broken behaviour as intended.

Separately `defineCommand: vi.fn()` (:111) is a no-op and `reclaimOrphan` (:158-190) re-implements the contract in JS, so RECLAIM_ORPHAN_SCRIPT (:89-112) has ZERO execution coverage. A wrong KEYS index, a swapped ARGV, or a `tonumber` failure on the `0.<timestamp>` resume score would pass every test in this file.

The tests do discriminate against develop (`reapOrphanClaims` does not exist there), so they are not worthless — but they validate the mock, not the shipped script. Marking blocking because the one blocking defect above is exactly what this suite was written to prevent and cannot see. Please add at least one test that executes the real Lua against a redis-mock/testcontainer, and change the mock's pending to key on the member string.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -163,6 +263,222 @@ export class RedisQueueAdapter implements QueueManager {
...
+
+            const enqueuedAtMs = Date.parse(parsed.enqueuedAt);
+            if (Number.isNaN(enqueuedAtMs)) {
+              this.logger.warn(
+                { workerId, itemKey, enqueuedAt: parsed.enqueuedAt },
+                'Unparseable enqueuedAt in reapOrphanClaims, skipping',
+              );
+              continue;
+            }
+            const ageMs = now - enqueuedAtMs;
```
christrudelpw: The FR-005 grace window measures age-since-ENQUEUE, not age-since-CLAIM, so it is inert for any item that waited in pending — it provides no mid-startup protection at all.

**Failure scenario:** `const ageMs = now - enqueuedAtMs;` where `enqueuedAt` is stamped by the monitor when the QueueItem is built, copied verbatim by CLAIM_SCRIPT, and preserved again by the reclaim. Nothing records a claim timestamp anywhere. So the Lua guard `if tonumber(ARGV[2]) < tonumber(ARGV[3])` with a 30s window only protects items ENQUEUED under 30s ago — never items CLAIMED under 30s ago, which is the case it exists for.

Real protection comes solely from CLAIM_SCRIPT setting the heartbeat atomically at claim time. But that TTL is hardcoded to 30s and `startHeartbeat` runs only AFTER `requestLease`, whose `lease.requestTimeoutMs` default is also exactly 30000. With `lease.enforce=true` against a cloud that never answers (the documented fail-open case) the claim heartbeat expires at the same instant the lease request times out. A reaper cycle landing in that window sees EXISTS heartbeat = 0 and ageMs >> 30s, reclaims — and then `pollOnce` fails open and dispatches the worker anyway. The item runs while also sitting in pending.

Fix: record `claimedAt` in CLAIM_SCRIPT and grace-gate on that.

packages/orchestrator/src/services/drop-log-helper.ts
```diff
@@ -0,0 +1,111 @@
...
+ * same wedged itemKey emit at `info` so log queries can still surface
+ * them but alerting stays quiet.
+ */
+export function emitDropLog(
+  logger: Logger,
+  decision: DropSeverityDecision,
+  payload: Record<string, unknown>,
+  message: string,
+): void {
+  if (decision.severity === 'warn' && decision.isTransitionEdge) {
```
christrudelpw: The 'one warn when it clears' half of the transition-edge contract is not implemented — the clearing edge emits at `info`, so an operator sees the wedge open and never sees it close.

**Failure scenario:** The docstring at :37-39 promises 'Emits `warn` **once** when the entry clears (drops back below threshold)', and the approved answer asked for 'One line when an entry crosses the threshold, one when it clears'. But `emitDropLog` gates on `decision.severity === 'warn' && decision.isTransitionEdge` — on the warn→info edge the severity is `'info'`, so it falls through to `logger.info` and is indistinguishable from the noise it replaced.

FR-008 names 'an operator paging on warns' as the audience. They get paged on the open and never learn it resolved, so every wedge stays open in their mental model until they check by hand. The code also contradicts the comment a future reviewer would rely on.

packages/orchestrator/src/services/worker-dispatcher.ts
```diff
@@ -572,6 +572,33 @@ export class WorkerDispatcher {
       } catch (error) {
         this.logger.error({ err: error }, 'Error during reaper cycle');
       }
+
+      // #1054 / FR-001 / AD-4: sequential Redis-side reap after the in-memory
+      // pass, so the Redis sweep sees a coherent state. Error-tolerant (R2/R3)
+      // — a Redis error must not skip subsequent cycles.
+      const report = await this.queue
```
christrudelpw: The new reap call sits OUTSIDE the reaper loop's try/catch, so a synchronous throw permanently kills the reaper — including the pre-existing `reapStaleWorkers`.

**Failure scenario:** `await this.reapStaleWorkers()` is wrapped at :574-577, but `const report = await this.queue.reapOrphanClaims().catch(...)` sits after the closing brace. `.catch()` handles rejections only. If `this.queue.reapOrphanClaims` is undefined — any hand-rolled `as unknown as QueueManager` mock, or an older adapter instance in a mixed deploy — the property access throws a TypeError synchronously, the `while (!signal.aborted)` loop exits, the `Promise.all([pollLoop, reaperLoop])` rejects, and stale-worker reaping stops for the process lifetime while polling keeps running.

Same shape as the round-3 finding on PR #1052: a guard that protects the settle but not the invocation. Moving the call inside the existing try block costs nothing.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -163,6 +263,222 @@ export class RedisQueueAdapter implements QueueManager {
...
+   *
+   * Race safety (US2): the outer-loop `EXISTS` check is an optimization;
+   * the load-bearing guard is the server-side re-check inside the Lua
+   * script. Grace window (FR-005): claims younger than
+   * `2 × heartbeatCheckIntervalMs` are skipped.
+   *
+   * Never throws. On transport error mid-sweep: `warn` + return the
+   * partial report so subsequent cycles retry.
+   */
+  async reapOrphanClaims(now: number = Date.now()): Promise<ReapReport> {
```
christrudelpw, line 397: The sweep is driven by `claimed:*` keys, so an in-flight member whose backing claim hash is gone remains unreclaimable — the residual half of the original wedge.

**Failure scenario:** `reapOrphanClaims` enumerates only `SCAN MATCH orchestrator:queue:claimed:*`; nothing reconciles `in-flight-items` members against `pending ∪ claimed`.

To be clear, the chosen direction is RIGHT for the reported incident: the sweep never reads heartbeats to find candidates, only to veto them, so it correctly fires when the heartbeat key is fully ABSENT rather than merely stale. That closes the original blind spot.

The residue: claim hashes carry ttl=-1, and this incident already shows untTL'd keys surviving indefinitely. If `claimed:<workerId>` is ever evicted or HDEL'd without the SREM while the single `in-flight-items` set key survives, the itemKey is wedged in-flight forever with no claim hash for the sweep to find — and every enqueue drops exactly as in #1054, with no reclaim path at all. Worth a periodic set-vs-(pending ∪ claimed) reconciliation, or a follow-up issue.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -138,8 +225,21 @@ export class RedisQueueAdapter implements QueueManager {
       } else {
         // #879 / FR-009: structured drop signal for the in-flight-collision path.
         // Distinct from the Redis-error warn path below.
-        this.logger.info(
-          { itemKey, reason: 'in-flight' },
+        // #1054 / FR-006: severity escalates from `info` to `warn` on the
+        // transition edge when the wedged in-flight entry's age crosses
+        // `maxRunDurationMs`. Structured fields preserved verbatim so
+        // downstream log queries and alerts key on unchanged shape.
+        const ageMs = await this.hasInFlightAge(itemKey);
```
christrudelpw, line 265: `hasInFlightAge` adds a full `SCAN claimed:*` plus an HGET per claimed key to every DROPPED enqueue — which the incident data shows is the steady-state hot path, not the rare path the comment claims.

**Failure scenario:** `const ageMs = await this.hasInFlightAge(itemKey);` runs inside `enqueueIfAbsent`'s collision branch, and the comment at :270-273 justifies the O(workers) cost with 'the drop path is not hot (fires only on collision)'.

The incident contradicts that directly: four monitors re-enqueue every in-flight issue on every poll cycle and each is dropped — that is what produced 17 drop lines for one item in 84 minutes. So each monitor cycle now costs (in-flight items × (1 SCAN round trip + 1 HGET per claimed key)) extra Redis calls, per monitor, forever — added load on the dedupe gate that every dispatch depends on.

This is the same per-cycle-amplification objection raised on #1050 finding 4 and honoured there. Cache the age per sweep, or read it from the reaper's existing HGETALL rather than re-scanning on the collision path.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -163,6 +263,222 @@ export class RedisQueueAdapter implements QueueManager {
...
+      do {
+        const [nextCursor, keys] = await this.redis.scan(
+          cursor,
+          'MATCH',
+          `${CLAIMED_KEY_PREFIX}*`,
+          'COUNT',
+          100,
+        );
+        cursor = nextCursor;
+        for (const key of keys) {
```
christrudelpw, line 353: `hasInFlightAge` scans only claimed hashes on Redis while the in-memory version also scans pending, so Q5=C's parity goal is broken and pending-side wedges never escalate.

**Failure scenario:** The Redis implementation loops `SCAN claimed:*` + `HGET key itemKey` and returns null when no claim hash holds the itemKey (:288-301). `InMemoryQueueAdapter.hasInFlightAge` (:154-172) checks `this.claimed` AND `this.pending`.

So a test written against the in-memory adapter for an item sitting in pending observes a real ageMs and escalates; the identical scenario on Redis returns null, which `classifyDropSeverity` treats as fail-safe `info` and — by design at drop-log-helper.ts:64-70 — does not even update the transition state.

Reachable in production: after a reclaim (or a `release()` retry) the item IS pending, and if dispatch is stalled (worker cap, or `pausePolling()` after a lease denial) it sits there indefinitely while every monitor cycle logs `info` with `ageMs: null` — the same invisible-wedge shape #1054 set out to eliminate, just one queue over.

christrudelpw (commented):
Review of generacy-ai/generacy#1056 — 9 finding(s), 3 blocking.

8 anchored inline; 1 could not be anchored to a diff hunk.


## General findings (no file anchor)

### Finding 1
**Files:** packages/orchestrator/src/services/redis-queue-adapter.ts
**Finding:** **BLOCKING** `packages/orchestrator/src/services/redis-queue-adapter.ts`:566 — Q3=C is satisfied literally — the reaper never dead-letters — but the inflated `attemptCount` it persists is consumed by `release()`'s dead-letter gate, so cluster restarts still condemn innocent items. That is precisely the outcome Q3=C was chosen to prevent.
**Failure scenario:** The reaper writes `attemptCount: attemptCountAfter` into the re-enqueued payload (:406-412) and never dead-letters — verified, the only DEAD_LETTER_KEY write is in `release()`. But `release()` READS that persisted count:

    attemptCount = parsed.attemptCount + 1;        // :563
    if (attemptCount >= this.maxRetries) { ... }    // :566, default 3

and `release()` is invoked by the pre-existing heartbeat-expiry path (`worker-dispatcher.ts:631`) — another infrastructure event.

So: three cluster restarts walk an untouched item to attemptCount=3, and the next heartbeat blip dead-letters it at 4, silently, without the item ever having failed. This cluster restarted twice today alone, which is exactly the frequency that makes this bite.

The approved Q3=C answer chose 'increment but never dead-letter from the reaper' specifically so infra events could not condemn blameless items — the observed orphan was caused by a restart at 19:22:48, five minutes after the claim, and had nothing to do with the item. Routing the counter into a gate that DOES dead-letter reintroduces the harm through the back door.

Fix: if the counter is diagnostic per Q3=C's rationale, make it a DISTINCT field (e.g. `reclaimCount`) that `release()` does not gate on. Keep `attemptCount` meaning 'this item's own execution failed'.
_reason: outside-diff-hunks_

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.orphan-reclaim.test.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/drop-log-helper.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/worker-dispatcher.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in 3cbe8df0 — please review, and re-open this thread if it still falls short.

PR #1064, opened 2026-07-28
feat: #1063 1063-problem-generacy-cloud-887

christrudelpw:
## Summary

Closes #1063

---


christrudelpw (commented):
Review of generacy-ai/generacy#1064 — 1 finding(s), 1 blocking.

packages/cluster-relay/src/messages.ts
```diff
@@ -357,6 +371,27 @@ const TierInfoMessageSchema = z.object({
...
+//   wroteDoc:  'created' | 'rebound'
+// .passthrough() preserves unknown top-level fields (Q1=A / FR-002) so future
+// cloud additions do not require a coordinated cluster release.
+const ClusterCockpitReplyMessageSchema = z
+  .object({
+    type: z.literal('cluster.cockpit.reply'),
+    timestamp: z.string(),
+    frameId: z.string().nullable(),
+    frameType: z.string(),
+    gateId: z.string().min(1),
```
christrudelpw: **BLOCKING** `gateId: z.string().min(1)` rejects the empty gateId the cloud deliberately sends on the unknown-subtype reply, so that reply fails to parse and produces `Invalid relay message, skipping` — this issue's exact bug, surviving on the one path that matters most during a rollout.

**Failure scenario:** The merged cloud code (generacy-cloud `develop`, `message-handler.ts`, from #890 / `a6944bf4`) builds the unknown-subtype reply with an explicitly best-effort, possibly-empty gateId:

    const bestEffortGateId = typeof data.gateId === 'string' ? (data.gateId as string) : '';
    const reply: ClusterCockpitReplyMessage = {
      type: 'cluster.cockpit.reply',
      frameId,
      accepted: false,
      frameType: 'unknown',
      gateId: bestEffortGateId,   // <-- '' when the frame carried none
      ...
    };

Its own tests pin this (`message-handler.test.ts` asserts `gateId: ''` for the no-gateId case).

`.min(1)` rejects `''`, so `RelayMessageSchema.safeParse` fails, `parseRelayMessage` returns null, and `relay.ts` logs `Invalid relay message, skipping`. That is precisely the symptom this issue exists to remove — reached via a field constraint instead of a missing union member.

And it lands on the worst possible path. The unknown-subtype reply exists specifically for the newer-cluster / older-cloud rollout window, or for a cloud that adds a subtype this cluster does not know. So the message designed to keep that window observable is the one message the cluster will drop and mislabel.

This is the same class as the Q2 answer: an over-strict validator on a field the sender treats loosely. Q2 loosened three enums for exactly this reason; `.min(1)` is the same mistake in a different shape, and it slipped through because the spec talked about enums rather than about constraints generally.

Fix: `gateId: z.string()` — drop `.min(1)`. Consider also asserting the empty-gateId case in `messages.test.ts`, since the cloud has a test pinning that it sends one.

Everything else in this PR matches the approved answers and is correct: `.passthrough()` (Q1=A), all three of `frameType` / `reason` / `wroteDoc` as `z.string()` (Q2=A), the structural `return` before the handler fanout with a comment explaining why (Q3=A), and the full parsed message at `debug` on the happy path with FR-005's enumerated set at `info` on `accepted: false` (Q4=C).

christrudelpw: Addressed in 2e1512b6 — please review, and re-open this thread if it still falls short.

christrudelpw (commented):
Round 2 — blocking finding resolved. Approving (posted as COMMENT; GitHub refuses APPROVE on a self-authored PR).

`gateId: z.string().min(1)` → `z.string()`, with a comment naming *why* the empty string is legitimate rather than just asserting it, plus a regression test that pins the cloud's unknown-subtype reply shape. That test is the part that matters: it makes the constraint un-reintroducible by someone tightening the schema for tidiness, because the test names the sender and the scenario.

Verified against the merged cloud source (`services/api/src/services/relay/message-handler.ts:831`):

```ts
const bestEffortGateId = typeof data.gateId === 'string' ? (data.gateId as string) : '';
```

The rest of the PR is unchanged and still matches the approved clarification answers: `.passthrough()` (Q1=A), all three tolerance fields as bare `z.string()` (Q2=A), the structural `return` before handler fanout in `relay.ts` (Q3=A), full message at `debug` with the enumerated drop set at `info` (Q4=C).

**Non-blocking, test fidelity.** The new fixture sends `reason: 'unknown-subtype'`, but the cloud sends `reason: 'invalid-payload'` on this exact path. Functionally irrelevant — `reason` is an open string by design — but the test comment cites the cloud handler as its source, so a future reader diffing the two will find a field that doesn't match. Worth aligning the literal next time this file is touched; not worth a round on its own.

One note for the rest of the #1059 wiring: the `.min(1)` slip is the same class of defect as the one Q2 was written to prevent — an over-strict validator on a field the sender treats loosely. Q2 framed tolerance in terms of *enums*, so a length constraint on a plain string read as unrelated. When reviewing the remaining union members, check every constraint against what the sender actually emits, not just the enum-shaped ones.

PR #1065, opened 2026-07-28
feat: #1060 1060-problem-redisqueueadapter

christrudelpw:
## Summary

Closes #1060

---


packages/orchestrator/src/services/worker-dispatcher.ts
```diff
@@ -257,24 +257,20 @@ export class WorkerDispatcher {
...
+    // its in-flight-SET membership (via `HDEL claimed` / `DEL heartbeat` /
+    // `ZADD pending` with no `SREM`). After the invariant fix, the itemKey
+    // is still in the in-flight SET after CLAIM, so a naked `enqueue()`
+    // here would be dropped by the new dedupe and leave the item
+    // orphan-claimed. Priority-tier divergence (release re-pends at `retry`,
+    // not `resume`): acceptable per research.md § Decision 3 Option R1.
     const itemKey = `${worker.item.owner}/${worker.item.repo}#${worker.item.issueNumber}`;
-    const isDuplicate = existingItems.some(
-      (qi) => `${qi.item.owner}/${qi.item.repo}#${qi.item.issueNumber}` === itemKey,
+    await this.queue.release(targetWorkerId, worker.item);
```
christrudelpw: **BLOCKING** `release()` has no in-flight/pending guard, and this PR deletes the `getQueueItems` duplicate check that was covering that gap — so the lease-expiry path can still produce two pending members for one itemKey, which is this issue's headline bug.

**Failure scenario:** The fix restores the invariant on `enqueue()` and then routes the lease-expiry path away from `enqueue()` to `release()`. But `release()` re-pends with a raw, unguarded `ZADD` (`redis-queue-adapter.ts:692`) — it never consults the in-flight SET and never checks whether a pending member for that itemKey already exists. Its else-branch runs whenever `attemptCount < maxRetries`, including when `hget(claimedKey, itemKey)` returned null.

Concrete sequence, with #1056's orphan reaper (merged `afab7d58`) as the concurrent actor:

1. Worker A claims `owner/repo#N`. In-flight SET holds the itemKey; claimed hash holds the payload.
2. Worker A dies. Its heartbeat key expires.
3. `reapOrphanClaims` sweeps it. `RECLAIM_ORPHAN_SCRIPT` does `HDEL claimed` + `ZADD pending <reclaim-payload>` and deliberately omits `SREM`, so the in-flight member survives. **Pending now has one member.**
4. The lease manager independently observes the same dead worker and fires `lease:expired` → `handleLeaseExpired` → `await this.queue.release(targetWorkerId, worker.item)`.
5. `release()` runs `hget(claimedKey, itemKey)` → **null** (the reaper already HDEL'd it) → `attemptCount = 0` → `0 >= 3` is false → else-branch → `ZADD pending <retry-payload>`.
6. The reaper's payload and release()'s payload differ in `queueReason` / `priority` / `attemptCount` / `enqueuedAt`. Redis ZSETs key on the member string, so these are **two distinct members for one itemKey**.
7. Two `CLAIM_SCRIPT` calls ZPOPMIN both and HSET them under two different workers → **two concurrent workers on one issue, conflicting pushes to one branch.**

That is step-for-step the failure sequence in this issue's own description, reached through `release()` instead of `enqueue()`.

The orphan reaper is not required to trigger this — it is just the most likely concurrent actor, since it and the lease manager are two independent detectors of the same event (a dead worker), so both firing is the expected case rather than an edge case. Any path that re-pends an item whose claim has already been cleared reaches the same place.

What makes this a regression rather than a pre-existing hole: the code being deleted here was the guard. The old `handleLeaseExpired` scanned `getQueueItems(0, 100)` for a pending entry matching `owner/repo#N` and skipped the re-enqueue when it found one. That check is exactly what stopped the duplicate member from being created. Replacing it with an unguarded `release()` removes the protection while the PR's stated purpose is to add protection.

Fix: `release()`'s retry branch needs the same in-flight/pending discipline `enqueue()` just gained — either a Lua guard that refuses to `ZADD` when a pending member for the itemKey already exists, or an explicit early return when `claimedRaw` is null (nothing was claimed, so there is nothing to release, and some other actor has already re-pended). The second is smaller and directly closes the sequence above. Whichever you pick, please pin it with a test that runs the reaper and the lease-expiry handler against one claimed item and asserts `ZCARD pending === 1`.

packages/orchestrator/src/services/worker-dispatcher.ts
```diff
@@ -257,24 +257,20 @@ export class WorkerDispatcher {
...
+    this.logger.info(
+      { itemKey, workerId: targetWorkerId },
+      'Re-enqueued via release() after lease expiry',
```
christrudelpw: **BLOCKING** Routing lease expiry through `release()` silently makes lease expiry consume a retry attempt and dead-letter the issue after 3 — a work-loss policy change that is in neither the spec, the plan note, nor the changeset, and that no test covers.

**Failure scenario:** Old path: `enqueue({...worker.item, priority: 0, queueReason: 'resume', enqueuedAt: now})`. No attempt increment, no dead-letter branch. A lease expiry could never terminate an issue.

New path: `release(targetWorkerId, worker.item)` → `attemptCount = parsed.attemptCount + 1` → `if (attemptCount >= this.maxRetries)` → dead-letter: `HDEL claimed`, `DEL heartbeat`, `ZADD dead-letter`, **`SREM in-flight`**. `maxRetries` defaults to 3 (`redis-queue-adapter.ts:195`).

So: an issue whose worker's lease expires three times over its lifetime is moved to `orchestrator:queue:dead-letter` and removed from the in-flight SET. It is never dispatched again. Its `agent:in-progress` label remains, its `process:` label was consumed at intake, and `markProcessed` already fired — so nothing re-enqueues it. The only signal is a single `warn` line. The issue stalls permanently, which is the shape of wedge #1054 was filed to eliminate.

The three expiries need not be consecutive or related: `attemptCount` is cumulative and persisted in the claimed-hash payload. Lease expiry fires on orchestrator restart, worker OOM-kill, container redeploy, and relay disconnect — none of which say anything about whether the work is failing. `maxRetries` is tuned for handler failures; this change makes one counter serve both, so two genuine handler failures plus one orchestrator restart dead-letters the issue.

The CLAUDE.md planning note discusses the priority-tier divergence (`retry` vs `resume`, accepted per research.md Decision 3 / R1) but never mentions the attempt increment or the dead-letter reachability, so this looks unintended rather than chosen.

Fix: pick one and say so explicitly. Either (a) re-pend without consuming an attempt on the lease-expiry path — infrastructure failure is not a work failure — via a `release()` variant or an explicit resume intent; or (b) accept the policy, document it in the spec and changeset, and add a test asserting the third lease expiry dead-letters. Right now no test exercises the dead-letter branch from lease expiry at all, so whichever behaviour you want is unpinned.

christrudelpw (commented):
Review of generacy-ai/generacy#1065 — 6 finding(s), 4 blocking.

6 anchored inline; 0 could not be anchored to a diff hunk.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.enqueue-invariant.test.ts
```diff
@@ -0,0 +1,479 @@
...
+
+/**
+ * #1060 — regression suite for the `RedisQueueAdapter.enqueue()` in-flight-SET
+ * invariant restoration. Uses the same stateful-mock harness pattern as
+ * `redis-queue-adapter.orphan-reclaim.test.ts` (ZSET keyed by full member
+ * string, not by itemKey — mirrors real Redis semantics so the double-enqueue
+ * hazard is not structurally hidden).
+ *
+ * The new `enqueueItem` Lua script (`ENQUEUE_SCRIPT`) is byte-identical to
+ * `ENQUEUE_IF_ABSENT_SCRIPT`, so the mock stub for both simulates the same
```
christrudelpw: **BLOCKING** The regression suite stubs `enqueueItem` with a hand-written TypeScript reimplementation of the Lua, so `ENQUEUE_SCRIPT` is never executed — the suite cannot fail for the class of bug it exists to pin, and CI already provides a real Redis that would run it.

**Failure scenario:** `createMockRedisWithState()` registers `enqueueItem: vi.fn(async (...) => { if (state.inFlight.has(itemKey)) return 0; state.inFlight.add(itemKey); state.pending.set(payload, ...); return 1; })`. That is a correct implementation of the intended semantics written in TypeScript. It is not `ENQUEUE_SCRIPT`.

Consequence: every assertion in the 479-line suite passes identically if `ENQUEUE_SCRIPT` is deleted, has `KEYS[1]` and `KEYS[2]` transposed, omits the `SADD` entirely, or is never wired into `defineCommand`. The mock supplies the right answer regardless of what the script says.

The bug being fixed is precisely 'the Redis command sequence on this path was wrong — a `ZADD` with no `SADD`'. A harness that substitutes its own command sequence for the real one cannot detect a mis-issued command sequence. The suite would have passed against the pre-fix code had the mock been written the same way.

The plan flagged this exact hazard and chose against it: CLAUDE.md, `#1060` section, Tests bullet — *'Real-Redis harness reuses the pattern from `redis-queue-adapter.orphan-reclaim.test.ts` (**not the manual mock** in `redis-queue-adapter.enqueueIfAbsent.test.ts`; mocked SET/HASH would not catch a mis-issued `SADD`/`SREM` command shape — the class of bug being fixed).'* The implementation went with the manual mock anyway. (The plan is also wrong that `orphan-reclaim.test.ts` is real-Redis — its own header says 'Reuses the stateful mock pattern from `redis-queue-adapter.enqueueIfAbsent.test.ts:16-125`. So the implementer followed the real precedent, not the plan's mischaracterisation of it. The precedent is still the wrong one to follow here.)

A real Redis is already available: `.github/workflows/ci.yml:59-64` runs a `redis:7` service with a `redis-cli ping` healthcheck. Nothing blocks executing the actual script.

Fix, in order of preference:
1. Point the FR-007 invariant sequence (`enqueue → claim → release-retry → reclaim-orphan → complete`) and the FR-008 wedge regression at the CI Redis service, so the real Lua runs and `SISMEMBER`/`SADD`/`ZADD` are exercised as issued.
2. If the suite must stay hermetic, add a static assertion over the script text and registration — that `ENQUEUE_SCRIPT` contains `SISMEMBER KEYS[2]`, `SADD KEYS[2]`, `ZADD KEYS[1]` in that order, and that `defineCommand('enqueueItem', ...)` is called with it. There is direct precedent: #1051's `repo-checkout.test.ts` (SC-005) statically asserts both fetch sites include `--prune` for exactly this reason.

The parity suite has the same property and needs the same treatment.

packages/orchestrator/src/services/label-monitor-service.ts
```diff
@@ -425,11 +425,18 @@ export class LabelMonitorService {
...
-    this.logger.info(
-      { owner, repo, issueNumber, command: queueItem.command, workflowName },
-      'Issue enqueued',
-    );
+    // #1060 / FR-003: `enqueue()` now returns a boolean and drops in-flight
+    // collisions internally (via emitDropLog). Observe the boolean, but
+    // proceed with post-enqueue steps regardless — a `false` return means
+    // the item was already in flight and the enqueue's intent is satisfied.
+    // No extra log line here; the adapter owns FR-005.
+    const enqueued = await this.queueManager.enqueue(queueItem);
```
christrudelpw: **BLOCKING** `enqueue()` returns `false` for both 'already in flight' and 'Redis transport error', and this caller's comment asserts `false` always means the intent was satisfied — then calls `markProcessed()`, so a Redis blip drops the issue permanently.

**Failure scenario:** The callee's own JSDoc (`types/monitor.ts:242`) is explicit: *'@returns true if enqueued, false if dropped (already in flight **or transport error**)'*. The `catch` at `redis-queue-adapter.ts:675` logs a `warn` and returns `false`.

This caller's comment says the opposite: *'a `false` return means the item was already in flight and the enqueue's intent is satisfied'*. The two statements are in the same PR and cannot both be true.

When the second one is wrong — a Redis timeout, a failover, a dropped connection during `enqueueItem` — the code proceeds past the `if (enqueued)` block to `await this.phaseTracker.markProcessed(owner, repo, issueNumber, parsedName)` on line 442. The issue is now marked processed for dedup, so the label monitor will not pick it up again, and it was never added to pending. It is silently dropped, with one `warn` line at the adapter as the only trace.

This behaviour is not itself new — the pre-PR `enqueue()` swallowed the same error and `markProcessed` ran regardless. But this PR is where the return value is introduced, and the entire reason to add a return value is to let the caller distinguish outcomes. It arrives having merged the one distinction that carries a work-loss consequence, and then documents the merge as an invariant a future caller will trust.

Fix: make `false` mean only 'already in flight'. Either rethrow from the `catch` and let the caller decide, or return a discriminated result (`'enqueued' | 'in-flight' | 'error'`). Then this caller can leave the issue unmarked on error so the next poll retries it. Please also correct the comment either way — as written it contradicts the method it is calling.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -215,6 +249,15 @@ export class RedisQueueAdapter implements QueueManager {
     this.enqueueIfAbsentCommandDefined = true;
   }
 
+  private ensureEnqueueCommand(): void {
+    if (this.enqueueCommandDefined) return;
+    this.redis.defineCommand('enqueueItem', {
+      numberOfKeys: 3,
```
christrudelpw: `enqueueItem` is registered with `numberOfKeys: 3` but the Lua body never references `KEYS[3]`; the unused `_dedup:<itemKey>` key is dead weight and makes the call CROSSSLOT-invalid under Redis Cluster.

**Failure scenario:** `ENQUEUE_SCRIPT`'s body touches only `KEYS[1]` (pending) and `KEYS[2]` (in-flight). `KEYS[3]` is passed as `${DEDUP_KEY_PREFIX}${itemKey}` and never read — the docblock says it is 'reserved for future D6-a upgrade'.

Two costs. First, declaring a key you do not touch defeats the point of the KEYS/ARGV split: under Redis Cluster, every declared key must hash to the same slot, and `orchestrator:queue:pending` / `orchestrator:queue:in-flight-items` / `orchestrator:queue:_dedup:owner/repo#N` do not — the call would fail CROSSSLOT. Standalone Redis (what runs today) does not check, so this sits latent until someone moves the orchestrator to a clustered Redis, at which point the whole intake path errors and `enqueue()` fail-safes to `false` for every item.

Second, the sibling `enqueueIfAbsent` registers the identical body with `numberOfKeys: 2` (line ~212 pre-PR). Two registrations of the same script with different key counts is the kind of asymmetry that reads as significant and is not.

Fix: `numberOfKeys: 2`, drop the third argument at the call site, and keep the docblock note that D6-a would add it back. Reserving an unused key costs more than adding one later.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -41,6 +42,38 @@ redis.call('ZADD', KEYS[1], tonumber(ARGV[2]), ARGV[3])
...
+ * KEYS[2] = in-flight SET
+ * KEYS[3] = `_dedup:<itemKey>` hash (reserved for future D6-a upgrade;
+ *           the D6-b body below does not touch it)
+ * ARGV[1] = itemKey
+ * ARGV[2] = priority (numeric string)
+ * ARGV[3] = serialized item JSON
+ *
+ * Returns 1 if enqueued, 0 if already in flight.
+ */
+const ENQUEUE_SCRIPT = `
```
christrudelpw: `ENQUEUE_SCRIPT` is a verbatim copy of `ENQUEUE_IF_ABSENT_SCRIPT`; two identical Lua sources must now be kept in sync by hand, and the new test's mock depends on that identity holding.

**Failure scenario:** The docblock states the script is 'byte-identical to `ENQUEUE_IF_ABSENT_SCRIPT`', and it is. Nothing enforces that. A future change to the dedupe semantics applied to one copy and not the other yields two verbs that silently disagree about what 'in flight' means — the same adapter-divergence shape this issue was filed about, just relocated from Redis-vs-in-memory to enqueue-vs-enqueueIfAbsent.

The new test compounds it: `redis-queue-adapter.enqueue-invariant.test.ts` stubs `enqueueItem` and `enqueueIfAbsent` with two separately-written but identical TypeScript bodies, on the stated assumption that the Lua sources match. If the sources diverge, the mocks keep agreeing and the suite keeps passing.

Given the bodies are identical and the key counts should be too (see the previous finding), `enqueue()` can just reuse the already-registered `enqueueIfAbsent` command — one script, one registration, the two verbs differing only in how the caller reads the boolean, which is what the docblock already says they do. If you prefer to keep them separate against a planned divergence, declare one shared `const` and register it twice so the identity is structural rather than a comment.

christrudelpw (commented):
Round 2 — all six findings resolved. Approving (posted as COMMENT; GitHub refuses APPROVE on a self-authored PR). CI green at job level on `79cb7888` — `ci` and `integration`, no failed steps.

Context for anyone reading the label history: the `blocked:stuck-feedback-loop` on #1060 was **not** a stuck loop. The fixer CLI hit its 20-minute timeout (`exitCode: 143`), and the partial-completion strategy pushed the work but recorded `success: false` / `reason: 'cli-did-not-complete'`, which routes to the no-diff disposition even though there was a 20-file diff. The code landed; only the thread bookkeeping was missing. I resolved the threads by hand rather than spend another 20-minute round on it. My review was the proximate cause — six findings with long bodies didn't fit the budget.

**F1 — reaper race.** Fixed better than I asked. Rather than guarding `release()` and leaving lease-expiry on it, the path moved to a new `requeueForResume()` verb that null-guards the claim read and returns early when `claimedRaw` is absent, so a reaper that already re-pended cannot get a second ZSET member added on top. `release()` gained the same guard.

**F2 — dead-letter on lease expiry.** `requeueForResume()` preserves `attemptCount` verbatim and re-pends at `resume` priority. Infrastructure events no longer consume a retry attempt, so three unrelated restarts can't dead-letter a blameless issue — and the priority tier is now right for the path, which the `release()` route had wrong.

**F3 — test validity.** `redis-queue-adapter.script-wiring.test.ts` plus `_ENQUEUE_IF_ABSENT_SCRIPT_FOR_TESTS` asserts the script text and its registration, so deleting or transposing the Lua now fails a test.

**F4 — transport-error conflation.** Resolved the right way: `enqueue()` throws on transport error instead of returning `false`, and `label-monitor-service.ts` catches it and returns without `markProcessed`, leaving the issue retryable on the next poll. `false` now means only "already in flight".

**F5/F6 — script duplication.** `ENQUEUE_SCRIPT` is gone; `enqueue()` reuses the existing `enqueueIfAbsent` registration at `numberOfKeys: 2`, which removes the unused `KEYS[3]` and the CROSSSLOT hazard along with the copy.

---

**Non-blocking follow-up, not for this PR.** `requeueForResume()` reads the claim with `HGET` and then re-pends in a separate `MULTI`. Between those two calls the reaper can still `HDEL` and re-pend, in which case the `HDEL` here is a no-op and the `ZADD` adds a second member — the same duplicate, through a much smaller window. The deterministic path this issue was filed for is closed; what remains is a millisecond-scale TOCTOU rather than a whole-handler one. Closing it properly means folding the read and the re-pend into one Lua script, the way `RECLAIM_ORPHAN_SCRIPT` already does. Happy to file it separately.

PR #1071, opened 2026-07-28
feat: #1069 1069-problem-requeueforresume

christrudelpw:
## Summary

Closes #1069

---


packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -153,6 +153,134 @@ redis.call('ZADD', KEYS[3], tonumber(ARGV[4]), ARGV[5])
...
+ */
+const RELEASE_SCRIPT = `
+local claimed = redis.call('HGET', KEYS[2], ARGV[1])
+if not claimed then
+  redis.call('DEL', KEYS[3])
+  return {0, -1}
+end
+
+local parsed = cjson.decode(claimed)
+local attemptCount = parsed.attemptCount + 1
```
christrudelpw: `local attemptCount = parsed.attemptCount + 1` aborts the whole script if the claim payload ever lacks `attemptCount`, and the failure mode is the orphan-claim wedge this line of work exists to eliminate. A `or 0` default is free insurance.

**Failure scenario:** Not reachable on any path I can find today — `enqueue()` seeds `attemptCount: 0`, `CLAIM_SCRIPT` re-encodes it, and `RECLAIM_ORPHAN_SCRIPT` preserves it — so this is defence-in-depth rather than a live bug. Raising it only because the consequence is disproportionate to the cost of preventing it.

In Lua, `nil + 1` raises rather than coercing. The old TypeScript did the same arithmetic where `undefined + 1` silently yields `NaN`, so a payload shape that previously degraded quietly now aborts the script. The abort surfaces as a Redis error, gets caught by `release()`'s `catch`, and is logged as `Redis error in release` — at which point **nothing has mutated**: the claim is still in the claimed hash, the heartbeat is gone or going, and no pending member was written. That is an orphan-claimed item with no live worker, which is precisely the generacy#1054 wedge, reached from the very method meant to unwind claims safely.

`REQUEUE_FOR_RESUME_SCRIPT` is milder: `base.attemptCount = parsed.attemptCount` on nil just drops the key, and `return {1, parsed.attemptCount}` truncates the Lua array so the caller destructures `attemptCount` as `undefined`. It is only used for a log line, so the item still re-pends correctly.

Suggested: `local attemptCount = (parsed.attemptCount or 0) + 1` in `RELEASE_SCRIPT`, and `base.attemptCount = parsed.attemptCount or 0` in `REQUEUE_FOR_RESUME_SCRIPT` for symmetry. Two tokens, and it converts a wedge into a benign restart of the retry count.

Worth considering because the payload is decoded from data that has already round-tripped through Redis, JSON, and Lua — the one place where assuming a field's presence is least defensible.

christrudelpw (commented):
Review of generacy-ai/generacy#1071 — **no blocking findings**. One non-blocking hardening note inline.

This is the cleanest of the round. Each answer landed as intended and two went further than I asked:

- **Q1=A** — `RELEASE_SCRIPT` dispatches both branches internally on `attemptCount >= maxRetries`, with `SREM` on the in-flight SET firing only on the dead-letter path. One round trip on both branches, and the caller still clears `dropLogState` / `enqueuedAtCache` on code 2, so the #1054 Map-growth bounds survive the move into Lua. That was the easiest thing to drop when relocating the mutation and it wasn't dropped.
- **Q2=A** — `cjson.decode` of both the claim and the caller's base payload, mutate, re-encode, all inside the script. `base.claimedAt = nil` correctly relies on Lua's delete-on-nil so the field is absent rather than null.
- **Q3=B** — `{code, attemptCount}` tuples, `{0, -1}` / `{1, N}` / `{2, N}`, switched on cleanly by the caller.
- **Q4=C** — the deterministic-interleave block exists and the file header documents the `HEAD~1` baseline procedure that demonstrates it FAILS pre-fix. That demonstration is the actual deliverable and I'm glad it's written down rather than assumed. The N=100 natural-race smoke test sits alongside it, and the round-trip count went into its own file as suggested.

Two things I checked specifically because they are how this kind of suite usually turns out vacuous:

- **`describeReal` skips only on an explicit `SKIP_REAL_REDIS_TESTS=1`** — not on Redis being unreachable. So in CI these either run or fail loudly; they cannot silently no-op. That is stronger than a skip-if-unavailable gate and it is the right call.
- **`flushdb` is DB-scoped and the four real-Redis files claim distinct DBs (12/13/14/15)**, so parallel vitest workers can't wipe each other, and a developer running these against a live dev-stack Redis won't clear the orchestrator's DB 0 queue. The comment explaining why `keyPrefix` was rejected — `SCAN`'s MATCH pattern isn't prefixed, which would break the reaper path the suite depends on — is exactly the sort of reasoning that saves the next person an hour.

The inline note is defence-in-depth on a nil `attemptCount`, not a live bug.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in c129ae61 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.requeueForResume-atomic.integration.test.ts
```diff
@@ -0,0 +1,263 @@
...
+// The env-provided URL (e.g. REDIS_URL=redis://host:6379) has no path
+// component, so we must append the DB explicitly rather than trust the
+// URL to carry it.
+const DB = 13;
+const REDIS_URL_BASE = process.env.REDIS_URL ?? 'redis://127.0.0.1:6379';
+const REDIS_URL = `${REDIS_URL_BASE.replace(/\/\d+$/, '')}/${DB}`;
+const skipReason = process.env.SKIP_REAL_REDIS_TESTS === '1'
+  ? 'skipped via SKIP_REAL_REDIS_TESTS=1'
+  : null;
+const describeReal = skipReason ? describe.skip : describe;
```
christrudelpw, line 50: **BLOCKING** The four real-Redis suites live under `src/services/__tests__/`, which the `ci` job runs — but the `redis:7` service is declared only on the `integration` job. All of them fail with ECONNREFUSED and burn ~2 minutes of CI on connection timeouts.

**Failure scenario:** Real CI failure on `c129ae61`, job `ci`, step `Test (packages)`:

```
❯ redis-queue-adapter.requeueForResume-atomic.test.ts   (3 tests | 3 failed)  31577ms
❯ redis-queue-adapter.release-atomic.test.ts            (5 tests | 5 failed)  52620ms
❯ redis-queue-adapter.round-trip-count.test.ts          (3 tests | 3 failed)  31579ms
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6379
```

`.github/workflows/ci.yml` declares the `redis:7` service under the **`integration`** job only. The `ci` job runs `pnpm -r run --if-present test`, which picks up every orchestrator suite including these four, and nothing sets `SKIP_REAL_REDIS_TESTS=1` there.

The `describeReal` gate is right and should stay as it is: skipping only on an explicit env var means a suite can never silently no-op, which is the property that makes these tests worth having. The gate is not the problem — the wiring is. (I called this design out approvingly in the previous round without checking that the `ci` job actually had Redis. It doesn't, and this is the consequence.)

Note the cost even ignoring the red: ~115 seconds of the `ci` job spent waiting on connections that cannot succeed, on every run.

**Pick one:**

1. Add the same `redis:7` service block to the `ci` job. Simplest, and these suites are fast against a live Redis.
2. Set `SKIP_REAL_REDIS_TESTS=1` on the `ci` job's test step and let the `integration` job be the one that runs them. Keeps `ci` hermetic. **If you take this route, the `integration` job must actually run the orchestrator package's tests** — please verify that rather than assume it, because a skip in `ci` plus no coverage in `integration` means these suites never run anywhere, which is strictly worse than today.
3. Move the four files into whatever tree `integration` owns.

Option 1 is the least likely to leave a coverage hole. Whichever you choose, the acceptance check is the same: confirm on the PR run that these suites report **passed**, not skipped and not absent.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -153,6 +153,134 @@ redis.call('ZADD', KEYS[3], tonumber(ARGV[4]), ARGV[5])
...
+ */
+const RELEASE_SCRIPT = `
+local claimed = redis.call('HGET', KEYS[2], ARGV[1])
+if not claimed then
+  redis.call('DEL', KEYS[3])
+  return {0, -1}
+end
+
+local parsed = cjson.decode(claimed)
+local attemptCount = (parsed.attemptCount or 0) + 1
```
christrudelpw, line 251: **BLOCKING** `tests/unit/services/redis-queue-adapter.test.ts` still stubs the old `hget` + `multi` sequence for `release()`, so 11 of its tests fail now that the mutation moved into Lua. That file is in the second test tree and was missed.

**Failure scenario:** Real CI failure on `c129ae61`, job `ci`, step `Test (packages)`:

```
❯ tests/unit/services/redis-queue-adapter.test.ts  (48 tests | 11 failed)
  × release > should increment attemptCount and re-queue item
      → expected "spy" to be called with arguments: [ …(2) ]
        Number of calls: 0
  × constructor defaults > should default maxRetries to 3
      → Cannot read properties of undefined (reading '0')
```

These are stale doubles, not behavioural regressions. `release()` no longer issues `hget` followed by a `multi()` pipeline — it invokes the `releaseItem` defined command — so the `multi` spy records zero calls and assertions that index into `multi.mock.calls[0]` throw on `undefined`.

The cause is the split test tree: this PR thoroughly updated `packages/orchestrator/src/services/__tests__/` and left `packages/orchestrator/tests/unit/services/` untouched. Both trees are run by `pnpm -r test`, so the second one is not optional.

Affected: the whole `release` describe, `queue priority > should set retry priority on release re-queue`, both `constructor defaults` cases, and the two `#1060 PR #1065 review finding` cases for the reaper-race null-guard and `requeueForResume`.

**Please re-point the doubles at the new command surface rather than deleting the cases.** Several of them — particularly `should NOT re-pend to ZSET when claim hash is already gone (reaper HDEL raced)` and `should NEVER dead-letter, even at attemptCount that would normally dead-letter via release()` — encode exactly the invariants this PR is preserving. Losing them would remove the cheap hermetic guard on behaviour that otherwise only the real-Redis suites cover, and per finding 1 those currently do not run in `ci` at all.

While you are in there, check whether any of these overlap with the new `src/services/__tests__/` cases and are now genuinely redundant — but default to keeping both; a hermetic mirror of a real-Redis assertion is worth its runtime.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.script-wiring.test.ts
```diff
@@ -156,3 +158,227 @@ describe('RedisQueueAdapter — defineCommand wiring (#1060 script-wiring assert
...
+    );
+  });
+
+  it('SREM removes from KEYS[5] (in-flight SET) — FR-006 dead-letter branch only', () => {
+    expect(RELEASE_SCRIPT).toMatch(
+      /SREM'?\s*,\s*KEYS\[5\]\s*,\s*ARGV\[1\]/,
+    );
+  });
+
+  it('increments attemptCount by exactly one on the retry side (FR-004)', () => {
```
christrudelpw: **BLOCKING** The FR-004 script-text assertion still pins `attemptCount = parsed.attemptCount + 1`, which my requested `or 0` hardening changed to `(parsed.attemptCount or 0) + 1`. Re-pin the regex to the new text — do not relax it.

**Failure scenario:** Real CI failure on `c129ae61`, job `ci`, step `Test (packages)`:

```
× RELEASE_SCRIPT text (#1069 script-wiring assertions)
  > increments attemptCount by exactly one on the retry side (FR-004)
  → expected '\nlocal claimed = redis.call(\'HGET\'…' to match /attemptCount\s*=\s*parsed\.attemptCou…/
```

This one is on me: the `or 0` default I asked for in the previous round is correct, and it invalidated a pinning assertion I did not check for. Straightforward to fix, but worth fixing precisely.

The regex must be re-pinned to the new body — something that matches `(parsed.attemptCount or 0) + 1` while still proving the increment is exactly one. **Do not weaken it to something like `/attemptCount/`** to make it pass. The whole point of this assertion is that it fails when the arithmetic changes, and a fuzzy pattern would have silently tolerated the very edit that broke it — which would make it decorative rather than diagnostic.

A good shape keeps both facts pinned: the `or 0` default is present, and the increment is `+ 1`. Consider asserting on the normalized line rather than a loose substring, so a future change to either half still trips it.

The sibling FR-003 assertion in `REQUEUE_FOR_RESUME_SCRIPT` (`preserves parsed.attemptCount verbatim`) passed and should be checked too — that script also gained `or 0`, so if its assertion still passes, confirm it is matching for the right reason and not because it is already too loose.

christrudelpw (commented):
Round 2 — CI is red on `c129ae61`: **7 test files, 27 tests failing** in job `ci`, step `Test (packages)`. Three distinct causes, all inline.

My previous round approved this on a review of the diff. That was premature — the first round's CI never ran, so nothing had actually been built or tested when I said it looked clean. The Lua and the scripts still look right; what's broken is wiring and stale pinning around them.

1. The four real-Redis suites run in the `ci` job, which has **no** `redis:7` service — only `integration` declares one. 11 tests fail on ECONNREFUSED and ~115s per run is spent on connection timeouts. I praised the `describeReal` gate last round without checking that `ci` had Redis; the gate is correct, the wiring isn't.
2. `tests/unit/services/redis-queue-adapter.test.ts` (the other test tree) still stubs `hget` + `multi` for `release()`. 11 stale-double failures, not behavioural regressions.
3. The FR-004 script-text assertion pins the pre-`or 0` body — broken by the hardening I asked for last round. Re-pin it, don't relax it.

None of these touch the fix itself. Nothing in the review changes my read that the scripts, the tuple contract, the cache cleanup on the dead-letter path, and the deterministic-interleave harness are all right.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.requeueForResume-atomic.integration.test.ts
christrudelpw, line 50: **BLOCKING** Correction to my previous finding — the fix is a file rename to `*.integration.test.ts`, not a workflow edit. The repo already routes real-dependency suites to the Redis-having job by filename, and none of the three options I gave last round used that mechanism.

**Failure scenario:** My earlier finding offered three fixes and all three were wrong or unusable, which is very likely why last cycle produced no diff. Two required editing `.github/workflows/ci.yml` (a push that commonly fails without the `workflow` token scope), and the third said 'move into whatever tree integration owns' without naming it. Correcting that now with the actual mechanism.

The repo already separates these by **filename convention**, not by job configuration:

- `packages/orchestrator/vitest.config.ts` — `include: ['tests/**/*.test.ts', 'src/**/*.test.ts']`, `exclude: ['**/*.integration.test.ts']`
- root `vitest.integration.config.ts` — `include: ['**/*.integration.test.ts']`, `testTimeout: 30000`
- `ci` job runs `pnpm -r run test` → the orchestrator config → **excludes** `*.integration.test.ts`
- `integration` job runs `pnpm -r run test:integration` → the root integration config → **only** `*.integration.test.ts`, and this is the job that declares the `redis:7` service

So a suite that needs real Redis simply has to be named `*.integration.test.ts`. There is existing precedent in the repo (`cockpit-gates-integration.integration.test.ts`).

Note this also means my earlier option 2 would have been actively harmful: setting `SKIP_REAL_REDIS_TESTS=1` on the `ci` job would not have gotten these running in `integration` either, because `test:integration` never globs a plain `*.test.ts`. They would have run nowhere.

**Rename these four** (git mv, no content change needed):

- `redis-queue-adapter.requeueForResume-atomic.test.ts` → `redis-queue-adapter.requeueForResume-atomic.integration.test.ts`
- `redis-queue-adapter.release-atomic.test.ts` → `redis-queue-adapter.release-atomic.integration.test.ts`
- `redis-queue-adapter.round-trip-count.test.ts` → `redis-queue-adapter.round-trip-count.integration.test.ts`
- `redis-queue-adapter.attemptcount-preservation.test.ts` → `redis-queue-adapter.attemptcount-preservation.integration.test.ts`

Leave `orphan-reclaim`, `reclaim-lua`, `enqueue-invariant`, `enqueueIfAbsent` and `script-wiring` alone — those are mock-based and belong in the fast `ci` run.

Keep the `describeReal` / `SKIP_REAL_REDIS_TESTS` gate exactly as it is. It is correct: skipping only on an explicit opt-out means these can never silently no-op, which is the property that makes them worth having.

Acceptance: on the next PR run, the `ci` job no longer reports these files at all, and the `integration` job reports them **passed** — not skipped, not absent. Please confirm both halves; 'ci is green' alone would also be true if they simply stopped running anywhere.

christrudelpw (commented):
Round 3 — correcting my own finding, which is the likely reason last cycle produced no diff.

The fixer ran cleanly for 10 minutes (`exitCode: 0`) and changed nothing. Looking at why: my finding 1 offered three fixes and all three were unusable. Two required editing `.github/workflows/ci.yml`, which commonly fails without the `workflow` token scope; the third said 'move into whatever tree integration owns' without naming it. That is not an actionable instruction.

The repo already solves this by **filename convention** and I missed it: `vitest.config.ts` excludes `**/*.integration.test.ts`, the root `vitest.integration.config.ts` includes exactly those, and the `integration` job — the one with the `redis:7` service — runs `test:integration`. So the fix is a four-file `git mv`, no workflow change at all. Corrected finding inline with the exact renames.

Worth flagging that my earlier option 2 would have been actively harmful: `SKIP_REAL_REDIS_TESTS=1` on the `ci` job would not have gotten these into `integration` either, since `test:integration` never globs a plain `*.test.ts`. They would have run nowhere and CI would have gone green — the worst outcome available.

The other two findings from round 2 still stand as written and are unaffected by this correction:

- `tests/unit/services/redis-queue-adapter.test.ts` — 11 stale mock doubles still stubbing `hget` + `multi` for `release()`. Re-point them at the command surface; don't delete the reaper-race and never-dead-letter cases.
- `script-wiring.test.ts` FR-004 — re-pin the regex to the `(parsed.attemptCount or 0) + 1` body. Don't relax it.

I've cleared `blocked:stuck-feedback-loop` so this can be re-dispatched.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.requeueForResume-atomic.integration.test.ts
christrudelpw: Addressed in cc8bbb83 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in cc8bbb83 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.script-wiring.test.ts
christrudelpw: Addressed in cc8bbb83 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/__tests__/redis-queue-adapter.requeueForResume-atomic.integration.test.ts
christrudelpw: Addressed in cc8bbb83 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw, line 251: **BLOCKING** One stale mock left: `queue-adapter-parity.test.ts` stubs `enqueueIfAbsent` but not `releaseItem`, so `release()` throws into its own catch, mutates nothing, and the parity assertion fails. Same class as the file already fixed — this one was just missed.

**Failure scenario:** Real CI failure on `68b2269f`, job `ci`, step `Test (packages)` — the only remaining one, down from 27:

```
FAIL src/services/__tests__/queue-adapter-parity.test.ts
  > 'redis' adapter — enqueue() FR-007 parity (#1060 / SC-003)
  > release retry then complete clears in-flight only after complete
AssertionError: expected null not to be null
  ❯ src/services/__tests__/queue-adapter-parity.test.ts:238:27
```

The mock at `:86-91` defines `defineCommand: vi.fn()` and an `enqueueIfAbsent` stub, but nothing for `releaseItem` or `requeueForResumeItem`. So the chain is:

1. `release('worker-A', claimed)` calls `this.redis.releaseItem(...)` — `undefined` on this mock.
2. The resulting TypeError is swallowed by `release()`'s own `catch`, which logs `Redis error in release` and returns.
3. Nothing mutates: the item is still in worker-A's claimed hash and still absent from pending.
4. `claim('worker-B')` therefore returns `null`, and `expect(reclaimed).not.toBeNull()` fails at line 238.

Note the shape of this: the adapter's defensive `catch` turns a missing stub into a silent no-op rather than a loud failure, so the test fails two assertions downstream of the real cause. Worth remembering when reading the next one of these.

**Fix:** add `releaseItem` and `requeueForResumeItem` stubs to `createMockRedis` in `queue-adapter-parity.test.ts`, mirroring the Lua semantics — the same treatment already applied to `tests/unit/services/redis-queue-adapter.test.ts` in `4a006c61`. Both must return the `{code, attemptCount}` tuple shape:

- `requeueForResumeItem` → `{0, -1}` when the claim is absent, else `{1, attemptCount}` with the count preserved verbatim, moving the item claimed→pending and leaving in-flight membership intact.
- `releaseItem` → `{0, -1}` when absent; `{2, n}` when `attemptCount + 1 >= maxRetries` (claimed→dead-letter, **and** `SREM` in-flight); otherwise `{1, n}` (claimed→pending, in-flight preserved).

The dead-letter branch's `SREM` is the part to get right — this specific test asserts `hasInFlight` stays `true` across a release and only flips `false` after `complete()`, which is exactly the FR-007 invariant, so a stub that drops in-flight membership on the retry branch would pass the immediate assertion and silently break the one this test exists for.

Please also grep for any other mock defining `enqueueIfAbsent` without the two new commands, so this is the last round of this. `git grep -l "enqueueIfAbsent: vi.fn" packages/orchestrator` is the quick check.

christrudelpw (commented):
Round 4 — **one** failure left, down from 27. Everything from round 2 is confirmed fixed.

First, the reason CI had gone silent: PR was in merge conflict with develop (`CLAUDE.md` — #1069 and #1070 each appended a planning-notes section). GitHub can't build `refs/pull/N/merge` for a conflicted PR, so `pull_request` workflows never fire at all. That's why three pushes and a draft/ready toggle produced nothing, and why the last run predated #1070 landing on develop by 69 seconds. Not an outage, not a quota — I chased both. Resolved by keeping both sections, #1070's first.

Confirmed green on `68b2269f`:

- `integration` job **passed** — the four renamed `*.integration.test.ts` suites now run against the `redis:7` service, which is the half that matters. 'ci is green' would also have been true if they'd stopped running anywhere, so this is the check worth having.
- `tests/unit/services/redis-queue-adapter.test.ts` — all 11 stale doubles now pass.
- `script-wiring.test.ts` FR-004 — passes, and the re-pin is tight: it pins the `or 0` guard **and** the `+ 1` on the normalized line, with a comment explaining why a loose form would tolerate the exact edit it guards.

The remaining failure is one more mock that was missed in the same sweep. Inline.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in f81f9ffb — please review, and re-open this thread if it still falls short.

PR #1072, opened 2026-07-28
feat: #1070 1070-problem-when-pr-feedback

christrudelpw:
## Summary

Closes #1070

---


packages/orchestrator/src/services/pr-feedback-monitor-service.ts
```diff
@@ -370,6 +399,62 @@ export class PrFeedbackMonitorService {
       );
       issueLabels = [];
     }
+    // #1070 FR-006 / D-4: retry-eligible check fires BEFORE the blocked:*
+    // short-circuit. Preserves Assumption 5 — any UNRECOGNIZED blocked:*
+    // label still pauses the monitor; only the specific `blocked:fixer-timeout`
+    // (retry-eligible) is carved out here.
+    if (issueLabels.includes('blocked:fixer-timeout')) {
```
christrudelpw: **BLOCKING** The retry branch removes the label and consumes a retry from the budget BEFORE the generic `blocked:*` guard runs, so when any other handler's blocked label is also present it silently burns a retry, deletes the timeout signal, and dispatches nothing.

**Failure scenario:** The retry-eligible branch at `:406` mutates two pieces of state — it calls `client.removeLabels(...)` against GitHub and then `this.fixerTimeoutRetryCount.set(stateKey, priorRetries + 1)` at `:425` — and only afterwards does control reach the generic guard at `:458`:

```ts
const blockedLabel = issueLabels.find(l => l.startsWith('blocked:'));
if (blockedLabel) { /* skip, no dispatch */ }
```

So the budget is spent before anyone has established that a dispatch will actually happen. When a second `blocked:*` label is present, the guard correctly refuses to dispatch — but the mutations have already landed and are not undone.

This is reachable through ordinary operation, not operator error. Three independent handlers write `blocked:*` labels to the same issue:

- `pr-feedback-handler.ts:1042` → `blocked:stuck-feedback-loop` (and now the three timeout labels)
- `merge-conflict-handler.ts:703` → `blocked:stuck-merge-conflicts`
- `validate-fix-handler.ts:293` → `blocked:stuck-validate-fix`

They watch the same issue and neither coordinates with nor is gated by the others, so two coexisting blocked labels is a normal state.

Concrete sequence:

1. PR-feedback times out with a partial push. Handler applies `blocked:fixer-timeout`. Counter is 0.
2. Independently, the merge-conflict handler applies `blocked:stuck-merge-conflicts` to the same issue.
3. Monitor polls. `:406` matches. `priorRetries = 0 < 2`, so it removes `blocked:fixer-timeout` from the issue, sets the counter to 1, and filters the label out of the local snapshot.
4. `:458` finds `blocked:stuck-merge-conflicts` and skips. **No retry is dispatched.**
5. The `blocked:fixer-timeout` label is now gone from the issue, so `:406` will never match again. The timeout signal is permanently destroyed and one of the two retries is spent on nothing.
6. Once the merge conflict clears, the next normal dispatch carries `retryAttempt: 1`, so the PR gets one auto-retry instead of two — silently, with nothing in the log explaining why.

Step 5 is the part that matters most for this issue specifically: an operator looking at that issue now sees only `blocked:stuck-merge-conflicts` and has no way to know a fixer timeout ever occurred. That is exactly the class of signal-destruction #1070 exists to eliminate, reintroduced by the fix.

**Fix — reorder so the guard precedes the mutation.** Compute the other-blocked-label check first, over the label set minus `blocked:fixer-timeout`, and only enter the retry branch when nothing else is blocking:

```ts
const otherBlocked = issueLabels.find(
  l => l.startsWith('blocked:') && l !== 'blocked:fixer-timeout',
);
if (otherBlocked) { /* existing skip path */ return false; }
if (issueLabels.includes('blocked:fixer-timeout')) { /* retry branch, unchanged */ }
```

The general principle is worth applying to the whole branch rather than just this one case: **do not consume budget or mutate remote state until the dispatch is committed.** The same shape bites more weakly if the enqueue is later dropped as already-in-flight — the counter has advanced while the in-flight item still carries the older `retryAttempt`. Deferring both mutations until immediately before the enqueue closes that too.

Please add a test with two blocked labels present asserting that the counter is unchanged and `blocked:fixer-timeout` still on the issue after the poll.

christrudelpw (commented):
Review of generacy-ai/generacy#1072 — 1 blocking finding. Everything else matches the approved answers.

Kept deliberately short: this is the issue about the 20-minute fixer budget, and a long review is the thing most likely to trip the bug it fixes.

**What's right**, so the fixer doesn't churn on it:

- Q2=B — `SpawnClaudeResult` carries `exitCode` alongside `timedOut`, and the JSDoc's rule that callers must switch on `timedOut` rather than the numeric 143 is the correct invariant to write down.
- Q3=C — three distinct labels, and the `catch` in `spawnClaudeForFeedback` preserves `timedOut` rather than flattening it to false.
- Q4=A — `waiting-for:address-pr-feedback` survives every branch.
- Q5=C — budget arithmetic is consistent end to end: monitor dispatches on `priorRetries < 2`, handler terminates on `retryAttempt >= 2`, giving exactly two auto-retries, and the sole reset is the `totalUnresolvedThreads === 0` branch.
- The contradictory `msg: "Successfully pushed…" success: false` line is now structurally unreachable — split into two branches that each guard on `success`. That was the specific log defect in the report.
- Precedence placement is better than what I asked for: terminal timeout labels outrank `waiting-for:address-pr-feedback`, retry-eligible sorts below it because a retry is pending and "waiting" is the truer status. I hadn't considered the split and it's right.

The one finding is inline.

packages/orchestrator/src/services/pr-feedback-monitor-service.ts
christrudelpw: Addressed in f42b2153 — please review, and re-open this thread if it still falls short.

PR #1074, opened 2026-07-29
feat: #1066 1066-problem-generacy-cloud-890

christrudelpw:
## Summary

Closes #1066

---

### DIFF
--- a/.changeset/1066-frame-id-wire.md
+++ b/.changeset/1066-frame-id-wire.md
@@ -5,5 +5,11 @@
 
 Preserve caller-supplied `frameId` on `GateOpenSchema` / `GateOutcomeSchema` and
 the orchestrator cockpit-gates route so `cluster.cockpit.reply` correlation
-(generacy-cloud#890) stops collapsing onto `(gateId, frameType)` on idempotent
-retries. Additive-optional wire-schema field; older callers unaffected.
+(generacy-cloud#890) *can* stop collapsing onto `(gateId, frameType)` on
+idempotent retries. Additive-optional wire-schema field; older callers
+unaffected.
+
+This makes correlation *possible*, not delivered. Nothing on the cluster yet
+generates a per-frame `frameId` or keeps a `frameId → pending-promise` map, so
+outbound frames still omit the field and replies still carry `frameId: null`
+until a producer lands in a follow-up.
--- a/.changeset/1066-frame-id-wire.md
+++ b/.changeset/1066-frame-id-wire.md
@@ -1,0 +1,9 @@
+---
+"@generacy-ai/cockpit": minor
+"@generacy-ai/orchestrator": patch
+---
+
+Preserve caller-supplied `frameId` on `GateOpenSchema` / `GateOutcomeSchema` and
+the orchestrator cockpit-gates route so `cluster.cockpit.reply` correlation
+(generacy-cloud#890) stops collapsing onto `(gateId, frameType)` on idempotent
+retries. Additive-optional wire-schema field; older callers unaffected.
--- a/packages/cockpit/src/gates/schema.ts
+++ b/packages/cockpit/src/gates/schema.ts
@@ -67,6 +67,14 @@ export const GateOpenSchema = z.object({
   allowFreeText: z.boolean(),
   sessionId: z.string().min(1),
   askedAt: z.string().datetime(),
+  // #1066 — preserved for cloud per-frame reply correlation. Cloud reads at
+  // services/api/src/services/relay/message-handler.ts:804 (raw `data.frameId`,
+  // upstream of the frozen payload schema). Empty string is normalized to
+  // absent so two concurrent frames both bearing `""` cannot collide on the
+  // cluster's `frameId → pending-promise` map.
+  frameId: z
+    .union([z.string().min(1), z.literal('').transform(() => undefined)])
+    .optional(),
 });
 export type GateOpen = z.infer<typeof GateOpenSchema>;
 
@@ -80,6 +88,10 @@ export const GateOutcomeSchema = z.object({
   outcome: z.enum(['applied', 'superseded', 'failed']),
   detail: z.string().optional(),
   at: z.string().datetime(),
+  // Same shape and rationale as GateOpenSchema.frameId — #1066.
+  frameId: z
+    .union([z.string().min(1), z.literal('').transform(() => undefined)])
+    .optional(),
 });
 export type GateOutcome = z.infer<typeof GateOutcomeSchema>;
 
--- a/packages/cockpit/src/gates/schema.ts
+++ b/packages/cockpit/src/gates/schema.ts
@@ -69,11 +69,17 @@ export const GateOpenSchema = z.object({
   askedAt: z.string().datetime(),
   // #1066 — preserved for cloud per-frame reply correlation. Cloud reads at
   // services/api/src/services/relay/message-handler.ts:804 (raw `data.frameId`,
-  // upstream of the frozen payload schema). Empty string is normalized to
-  // absent so two concurrent frames both bearing `""` cannot collide on the
-  // cluster's `frameId → pending-promise` map.
+  // upstream of the frozen payload schema). Three encodings of "no correlation"
+  // are normalized to absent: omitted, `""`, and `null`. `null` matches the
+  // cloud's own vocabulary (see `packages/cluster-relay/src/messages.ts:154`
+  // + `:387` — reply `frameId: string | null`), so a caller doing
+  // `?? null` round-trips without a 400.
   frameId: z
-    .union([z.string().min(1), z.literal('').transform(() => undefined)])
+    .union([
+      z.string().min(1),
+      z.literal('').transform(() => undefined),
+      z.null().transform(() => undefined),
+    ])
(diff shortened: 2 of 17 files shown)


.changeset/1066-frame-id-wire.md
```diff
@@ -0,0 +1,9 @@
+---
+"@generacy-ai/cockpit": minor
+"@generacy-ai/orchestrator": patch
```
christrudelpw, line 3: After this merges, nothing on the cluster ever sets `frameId`, and no filed issue ever will — so replies keep carrying `frameId: null` and correlation stays inert, which is the condition this issue exists to end.

**Failure scenario:** The code here is correct and its scope is defensible — the spec is explicit that this preserves a *caller-supplied* `frameId` and is safe to land alone. My concern is one level up: what happens after it lands.

I checked for a producer:

```
grep -rn 'frameId' packages/generacy/src   # → zero hits
```

`cockpit_gate_open.ts` and `cockpit_gate_ack.ts` build the `POST /cockpit/gates[/:id/ack]` bodies and never include the field. So on merge day: outbound frames still omit `frameId`, the cloud still echoes `frameId: null`, and correlation still collapses onto `(gateId, frameType)` — which generacy-cloud#887 Q1 rejected precisely because idempotent retry of `gate-open` for one `gateId` is the designed pattern.

The part that makes this worth raising rather than shrugging at is that **no filed issue closes the gap.** Walking #1059's own step list:

| step | work | issue | touches frameId production? |
|---|---|---|---|
| 1–2 | schemas + route preserve `frameId` | **#1066 (this)** | no — preservation only |
| 3 | `cluster.cockpit.reply` into `RelayMessageSchema` | #1063 (merged `069536eb`) | no — makes the echo parseable |
| 4–7 | `runId` through status/list/query-client/cloud/agency | #1067, cloud#892, agency#469 | no — all runId |

Nothing generates a per-frame id, and nothing keeps the `frameId → pending-promise` map that would consume the echo. `gh search issues frameId --owner generacy-ai --state open` returns only #1066, #1059, #1068, cloud#892 — none of them.

That directly threatens two things I'd otherwise expect to hold:

1. **#1059's acceptance criterion** — *"A reply's `frameId` matches the frame the cluster sent, end to end, asserted over a real WebSocket rather than a `vi.fn()`"* — is unreachable by any currently-filed work. #1068 (the e2e verification issue) will hit this as a hard blocker rather than a check.
2. **`packages/cluster-relay/src/relay.ts:331`** still reads:

   > `// Observability-only until #1059 steps 4-7 wire the frameId correlation.`

   Per #1059's own list, steps 4–7 are the runId work and have nothing to do with `frameId`. After this PR that comment is not just stale — it actively points the next reader at four issues that will never wire it, which is how a gap like this survives another month.

**I am not asking this PR to grow.** Widening it would re-create the oversized-diff problem that trips the 20-minute fixer budget next door in #1073. Land it as-is. I'll file the producer issue (generate a `frameId` per outbound cockpit frame + keep the pending map + resolve on `cluster.cockpit.reply`) and fix the `relay.ts:331` comment there, so the trail points somewhere real.

One thing worth doing *here*, though, since it costs a sentence: the changeset currently reads as though correlation now works. It would be more honest as "makes correlation *possible*; still inert until a caller supplies the field (see <follow-up>)". A changeset that overstates delivery is what let this gap open in the first place — generacy-cloud#890 and generacy#1055 both shipped documenting their own inertness, and #1059 exists because nobody filed the issue those docs assumed.

packages/cockpit/src/gates/schema.ts
```diff
@@ -67,6 +67,14 @@ export const GateOpenSchema = z.object({
   allowFreeText: z.boolean(),
   sessionId: z.string().min(1),
   askedAt: z.string().datetime(),
+  // #1066 — preserved for cloud per-frame reply correlation. Cloud reads at
+  // services/api/src/services/relay/message-handler.ts:804 (raw `data.frameId`,
+  // upstream of the frozen payload schema). Empty string is normalized to
+  // absent so two concurrent frames both bearing `""` cannot collide on the
+  // cluster's `frameId → pending-promise` map.
+  frameId: z
```
christrudelpw, line 77: `frameId: null` is rejected with a 400 while `frameId: ""` is normalized to absent — and `null` is the encoding the cloud itself uses for "no frameId", so a caller round-tripping that type gets a hard failure on an optional field.

**Failure scenario:** The union accepts a non-empty string or `""` (→ absent), and `.optional()` accepts omission. Everything else fails both members, so `GateOpenSchema.parse({…, frameId: null})` throws and the route returns `400 VALIDATION`. The new tests pin this deliberately (`['null', null]` in the reject table), so it is a choice rather than an oversight — I just think it is the wrong one.

The asymmetry matters because of where `frameId` values come from. `packages/cluster-relay/src/messages.ts:154` and `:387` type the reply's field as:

```ts
frameId: string | null;          // :154
frameId: z.string().nullable(),  // :387
```

`null` is the cloud's own vocabulary for "no frameId". So the natural cluster-side type for a correlation id is `string | null`, and the natural call is:

```ts
await postGateOpen({ ...record, frameId: pendingFrame?.id ?? null });
```

That 400s. The caller has to learn that this one optional field wants omission specifically — not `null`, which is what the same concept is called on the way back, and not `""`, which is silently tolerated. Three encodings of nothing, two accepted, one fatal, and the fatal one is the one the adjacent schema uses.

The Q2=C rationale I gave for normalizing `""` was *one representation of "no correlation available"*. That argument applies to `null` at least as strongly — arguably more, since nobody writes `""` on purpose but `?? null` is idiomatic. FR-004's "not `null`, not `""`, absent" is a statement about the **outbound frame**, and normalizing satisfies it exactly as well as rejecting does: the wire still carries no key either way.

**Fix** — one union member, same transform:

```ts
frameId: z
  .union([
    z.string().min(1),
    z.literal('').transform(() => undefined),
    z.null().transform(() => undefined),
  ])
  .optional(),
```

and flip `['null', null]` from the reject table to the normalize-to-absent case in both schema tests and both route tests.

Non-blocking because no caller sends `frameId` yet (see my other comment), so nothing is broken today — but that is also exactly why it is cheap to change now, before the first producer is written against the stricter shape.

Separately, and purely to record that I checked rather than assumed: the tests are right to assert absence via `Object.hasOwn(JSON.parse(JSON.stringify(parsed)), 'frameId')` instead of on `parsed` directly. Zod's `mergeObjectSync` sets a key whenever `key in ctx.data`, so after parsing `{frameId: ''}` the object *does* carry an own `frameId: undefined` — the spec's "contains `frameId` only if the caller supplied a non-empty string" is not literally true of the parsed object, only of the serialized frame. Testing at the JSON boundary is what makes that distinction harmless, and whoever wrote these clearly knew it.

christrudelpw (commented):
Review of #1074 (issue #1066) — **2 findings, 0 blocking.** Both are inline.

The implementation matches the spec's recommended shape exactly, and the shape is right: because `cockpit-gates.ts:318` forwards `parsed` rather than `request.body`, adding `frameId` to the two schemas is precisely what stops it being stripped — no route source change needed, and correctly none attempted. The spec's warning against "start forwarding `request.body`" was heeded.

What I verified rather than assumed:

- **Retain/drain (spec item 3)** — `retained-cockpit-events.ts:74` passes `data: head.data` through unchanged, so `frameId` already survives replay. The new test pins it against future reshaping rather than claiming credit for a change. Correct call.
- **Wire-level absence** — the tests assert through `JSON.parse(JSON.stringify(parsed))`, not on `parsed`. That is the right boundary: Zod sets an own key whenever `key in ctx.data`, so parsing `{frameId: ''}` yields `frameId: undefined` as a present key, and only serialization drops it. Testing at the JSON boundary makes the distinction harmless. Noted in detail on the schema comment.
- **Route coverage** — both routes × {non-empty forwarded, omitted absent, `""` normalized, non-string 400}. No gaps found.
- **No regression risk from the new strictness** — `git grep frameId` on develop returns zero hits in `packages/cockpit/src`, `packages/orchestrator/src`, and `packages/generacy/src`, so no existing caller can start 400ing.

The two findings, in order of what I think matters:

1. **The producer gap** (on the changeset) — after this lands, nothing sets `frameId` and no filed issue ever will, so replies keep carrying `null`. Not this PR's job to fix; I'm filing the follow-up and will correct the misleading `relay.ts:331` comment there. The ask here is one honest sentence in the changeset.
2. **`null` → 400 while `""` → absent** (on `schema.ts:75`) — `null` is the cloud's own encoding for "no frameId" (`messages.ts:387`), so `?? null` is the idiomatic call that fails. One extra union member; cheapest to change now, before the first producer is written against the stricter shape.

Neither blocks. If you'd rather take only #2 and leave the changeset wording, say so and I'll merge on that.

.changeset/1066-frame-id-wire.md
christrudelpw: Addressed in df874b55 — please review, and re-open this thread if it still falls short.

packages/cockpit/src/gates/schema.ts
christrudelpw: Addressed in df874b55 — please review, and re-open this thread if it still falls short.

PR #1075, opened 2026-07-29
feat: #1073 1073-problem-when-pr-feedback

christrudelpw:
## Summary

Closes #1073

---


packages/orchestrator/src/worker/pr-feedback-handler.ts
```diff
@@ -623,12 +662,25 @@ export class PrFeedbackHandler {
...
-        this.logger.warn(
-          { prNumber, issueNumber, outcomes },
-          'commit pushed but resolve batch had zero successes — persisting trigger, entering blocked-stuck-feedback-loop disposition',
-        );
-        await this.addBlockedStuckFeedbackLoopLabel(github, owner, repo, issueNumber);
+        // #1073 FR-013: split by head-advance signal. A cycle where the branch
+        // HEAD moved (via CLI-self-commit OR handler-commit) but zero threads
+        // resolved is a GitHub-API-shaped failure, not a fixer wedge —
+        // `blocked:resolve-failed` names that distinctly.
+        const headAdvanced = cliSelfCommitted || hasChanges;
```
christrudelpw: **BLOCKING** `headAdvanced` is always true at this point, so the `blocked:stuck-feedback-loop` arm below it is unreachable. The code presents a two-way decision that can only ever go one way, and it silently removes that label from the zero-resolve path entirely.

**Failure scenario:** Trace the guard that controls entry to this block. At `:599`:

```ts
if (!cliSelfCommitted && (!success || !hasChanges)) { /* stuck-feedback-loop */ return; }
```

So reaching the happy path at 7b requires `cliSelfCommitted || (success && hasChanges)`. Then at `:669`:

```ts
const headAdvanced = cliSelfCommitted || hasChanges;
```

- If `cliSelfCommitted` is true → `headAdvanced` is true.
- Otherwise `success && hasChanges` held, so `hasChanges` is true → `headAdvanced` is true.

There is no path to `:669` on which `headAdvanced` is false. The `else` arm — the one that keeps the original `'commit pushed but resolve batch had zero successes … blocked-stuck-feedback-loop'` warn and calls `addBlockedStuckFeedbackLoopLabel` — cannot execute. I tried to construct a counterexample including `cliSelfCommitted && !hasChanges && !success` and could not; every route into 7b sets one of the two disjuncts.

Two consequences, and the second is the one that matters:

1. Dead code carrying a live-looking comment (`FR-006 tail`), which will read to the next person as a case that occurs. Someone will eventually 'fix' a bug in a branch that never runs.
2. **`blocked:stuck-feedback-loop` is now unreachable from the zero-resolve path altogether.** That may well be the intent of Q1=B — a resolve failure is not a stuck loop — but it is a real narrowing of that label's meaning, and right now it is expressed accidentally, by a condition that happens to be a tautology, rather than stated. Anyone auditing where `stuck-feedback-loop` can still originate has to re-derive this proof to find out.

**Fix:** drop the conditional and the dead arm. At this point in the flow the head has necessarily advanced, so say that directly:

```ts
if (resolveSuccesses === 0) {
  // Reaching 7b implies cliSelfCommitted || (success && hasChanges), so the
  // head has advanced by construction — a zero-resolve cycle here is always
  // a GitHub-side failure, never a fixer wedge.
  this.logger.warn({ prNumber, issueNumber, outcomes, preFixSha, postFixSha: postCliSha },
    'commit pushed but resolve batch had zero successes — entering blocked:resolve-failed disposition (#1073)');
  await this.addBlockedResolveFailedLabel(github, owner, repo, issueNumber);
  return;
}
```

If instead you believe there is a reachable `headAdvanced === false` case, please name it in a comment and add a test that exercises it — because the current arrangement asserts one exists without demonstrating it.

Either way, worth a line in the changeset noting that `blocked:stuck-feedback-loop` no longer originates from the zero-resolve path, since that changes what an operator seeing that label can conclude.

christrudelpw (commented):
Review of generacy-ai/generacy#1075 — 1 blocking, 1 non-blocking. Both inline.

All four approved answers landed correctly:

- **Q1=B** — `blocked:resolve-failed` exists with a docstring that names the *remediation* difference (check GitHub API responses, not fixer transcripts). That framing is the right one: the label earns its place because it changes what the operator does next.
- **Q3=A** — `source` is on **both** paths, including `source: 'handler'` added to the two existing lines. That symmetry is the half that's easy to skip and the half that makes the field queryable.
- **Q4=A** — `preFixSha` + `postFixSha` are in the payload, so the head-advance claim is auditable rather than asserted. That was my caveat and it's honoured.
- **Q2=A** — no CLI-side commit-policy changes; scope held.

Precedence placement is right too: `blocked:resolve-failed` outranks `waiting-for:address-pr-feedback` because — unlike `blocked:fixer-timeout` — there's no auto-retry window, so the cluster genuinely isn't still waiting. That's the correct distinction between the two, and it mirrors the reasoning #1070 used to sort its own three labels.

The blocking item is a tautology that makes one arm of the new branch unreachable.

packages/orchestrator/src/worker/pr-feedback-handler.ts
```diff
@@ -589,6 +611,23 @@ export class PrFeedbackHandler {
         return;
       }
 
+      // #1073 FR-007/FR-008/FR-008a: distinct info line when the CLI committed
+      // and pushed its own work. `preFixSha` + `postFixSha` in the payload make
+      // the head-advance claim auditable (research.md § D-2).
+      if (cliSelfCommitted && !hasChanges) {
```
christrudelpw: The `source: 'cli'` line is gated on `!hasChanges`, so a mixed cycle — CLI commits some work, handler commits the rest — logs only `source: 'handler'` and the self-commit becomes invisible, which is the exact visibility this issue exists to add.

**Failure scenario:** `if (cliSelfCommitted && !hasChanges)` emits the `disposition: 'cli-self-committed'` line only when the handler found nothing left to commit. But `cliSelfCommitted` and `hasChanges` are independent: the CLI can commit some of its work and leave other files in the working tree, which `commitAndPushChanges` then commits as a second commit. That cycle sets both flags.

In that case the only lines emitted are the handler's `'Successfully pushed changes to PR branch'` with `source: 'handler'`. An operator grepping `disposition: 'cli-self-committed'` — the taxonomy SC-003 establishes — sees nothing, and the reconstruction of what happened requires reading git history, which is precisely the archaeology this issue set out to eliminate.

Not blocking because the dispositional behaviour is already correct (the `!cliSelfCommitted &&` guard at `:599` fires on the right condition regardless), and because the mixed cycle is the less common shape. But the fix is small and worth taking now: drop the `&& !hasChanges` so the line reflects what the CLI did independently of what the handler did afterwards.

If both commits happened, that is arguably the *most* interesting cycle to have in the log, since it is the one where the two committers interleaved.

A `commitCount` or a `handlerCommitted: boolean` field alongside `source` would let a reader distinguish the three shapes — CLI-only, handler-only, both — from one grep.

packages/orchestrator/src/worker/pr-feedback-handler.ts
christrudelpw: Addressed in 611003ab — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/pr-feedback-handler.ts
christrudelpw: Addressed in 611003ab — please review, and re-open this thread if it still falls short.

PR #1076, opened 2026-07-29
feat: #1058 1058-follow-up-1054-pr

christrudelpw:
## Summary

Closes #1058

---


packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -281,6 +282,54 @@ return {1, attemptCount}
...
+ * between the client-side residue computation and the Lua invocation for
+ * this specific candidate.
+ *
+ * KEYS[1] = orchestrator:queue:in-flight-items
+ * ARGV[1] = itemKey
+ *
+ * Returns:
+ *   1 = reconciled (SREM fired)
+ *   0 = skipped-race-reappeared (item was gone before we could SREM, or a
+ *       concurrent `enqueueIfAbsent`/`enqueue` re-added it after our
```
christrudelpw: The script's JSDoc describes a race guard the code does not implement — a concurrent re-add makes `SISMEMBER` return 1, not 0, so the documented "either way, we do NOT SREM" is the opposite of what runs. The real safety argument is elsewhere and should be the one written down.

**Failure scenario:** The docblock makes two claims about `RECONCILE_IN_FLIGHT_SCRIPT`:

> *this script's role is limited to catching the narrow window where an item genuinely re-entered flight between the client-side residue computation and the Lua invocation*

> `0 = skipped-race-reappeared (item was gone before we could SREM, or a concurrent enqueueIfAbsent/enqueue re-added it after our client-side detection but before this Lua fired — either way, we do NOT SREM)`

Both are wrong, in different directions:

**1. A re-add produces `1`, not `0`.** If something SADDed the itemKey back between detection and the Lua, `SISMEMBER` returns **1**, the `if exists == 0` branch is skipped, and the script **SREMs it**. The stated "either way, we do NOT SREM" is precisely inverted for the case it names. Anyone relying on this comment to reason about whether the guard is sufficient is reasoning from a false premise.

**2. The guard adds no protection at all.** `SREM` on a non-member is already a no-op, so `SISMEMBER` + guarded `SREM` is behaviourally identical to a bare `SREM`. What the `SISMEMBER` buys is the *return value* — a distinguishable `0` for reporting. It is a reporting shim, not a race guard, and calling it one in the docblock is what makes the inversion above dangerous rather than merely untidy.

**The reason this is non-blocking is the third fact, and it is the one that deserves the comment:** the scenario cannot happen. I checked rather than assumed. Since #1060 (`fbcf85fb`), `enqueue()` and `enqueueIfAbsent()` both route through the *same* in-flight-guarded Lua — `redis-queue-adapter.ts:415` and `:786` invoke the identical `enqueueIfAbsent` command, and its `SADD` is gated on `SISMEMBER IN_FLIGHT_KEY` returning 0. Every residue candidate is by definition **present** in `IN_FLIGHT_KEY`, so both verbs drop rather than SADD for exactly the itemKeys this sweep operates on. There is no code path that can re-add a residue member during the window. I grepped for other `SADD` sites against `IN_FLIGHT_KEY` and found none — `RECLAIM_ORPHAN` re-pends without SADD, `REQUEUE_FOR_RESUME` only ZADDs, and `RELEASE` only SREMs on dead-letter.

So the safety of this sweep rests on two things, neither of which is the SISMEMBER:

- the two-sweep tracker (`reconcileTracker`), which is what makes snapshot-interleaving artifacts self-clear; and
- the invariant that **every** `SADD` path is in-flight-guarded, so a member that is present cannot be re-added.

That second invariant is load-bearing and currently undocumented. If someone later adds an unguarded `SADD` — a bulk-requeue admin verb, say — this sweep silently becomes able to `SREM` a live item, and the comment will still be telling them the script handles it.

**Fix:**
- Rewrite the docblock to state the two real guarantees above, and describe the script honestly as "atomic remove + presence-report; not a race guard".
- Add the SADD-guard invariant as an explicit note, so a future unguarded SADD has something to trip over in review.
- Rename `skippedRaceReappeared` (or at minimum re-comment it). Result `0` means the member **disappeared** between snapshot and Lua — the inverse of "reappeared". `skippedAlreadyGone` says what it counts. Note this field name is inherited from `ReapReport`, where it *is* accurate, so the copy is understandable — but the two now mean opposite things under the same name, in reports logged one after the other by the same loop.

packages/orchestrator/src/services/worker-dispatcher.ts
```diff
@@ -105,6 +105,15 @@ export class WorkerDispatcher {
       'Starting worker dispatcher',
     );
 
+    // #1058 / AD-4 / Q2=B — boot sweep for `IN_FLIGHT_KEY` residue armed at
+    // process start. Fire-and-forget so `start()` returns immediately; the
+    // reaperLoop's first regular sweep runs its own tracker check regardless.
+    // Pre-crash residue armed by boot-sweep completion is repaired at the
+    // first regular sweep (~heartbeatCheckIntervalMs).
+    void this.queue.reconcileInFlight().catch((err) => {
```
christrudelpw: The boot sweep's `void queue.reconcileInFlight().catch(...)` is not inside a try/catch, so a synchronous throw takes down `start()` itself — the exact hazard #1054 finding 5 documents 25 lines below, with a larger blast radius.

**Failure scenario:** `.catch()` only handles a *rejected promise*. If `this.queue.reconcileInFlight` is `undefined`, the call throws `TypeError: this.queue.reconcileInFlight is not a function` **synchronously**, before `.catch()` is ever attached. In `start()` that rejection propagates out of `start()` — so `Promise.all([this.pollLoop(...), this.reaperLoop(...)])` on the next line is never reached and the dispatcher does not start at all. No poll loop, no reaper loop, no queue processing.

The reaper-loop call site three hundred lines down gets this exactly right, and says why:

> *#1054 finding 5 — the invocation MUST sit inside the surrounding try/catch, not outside. `.catch()` handles rejections only; a synchronous throw (undefined `reapOrphanClaims` on a hand-rolled `as unknown as QueueManager` mock, or an older adapter instance in a mixed deploy) would exit the `while (!signal.aborted)` loop and permanently kill BOTH reap paths for the process lifetime.*

The new `reconcileInFlight` call inside `reaperLoop` correctly sits within that try/catch and inherits the protection. The boot sweep does not, and its failure mode is strictly worse: finding 5's hazard killed the reap paths, this one kills the dispatcher.

**Not blocking, because I checked reachability and it is closed today.** `grep 'implements QueueManager'` returns exactly two classes — `RedisQueueAdapter` and `InMemoryQueueAdapter` — and this PR adds `reconcileInFlight` to both. The mixed-deploy variant finding 5 worried about does not apply here either, since both adapters ship in the same package as the dispatcher. The residual exposure is test doubles: `worker-dispatcher.terminal.test.ts` builds a dispatcher from an `as unknown as QueueManager` mock, which is precisely the shape finding 5 named.

So this is about keeping the defensive posture consistent with the reasoning the file already commits to, at one line of cost:

```ts
try {
  void this.queue.reconcileInFlight().catch((err) => {
    this.logger.warn({ err }, 'boot reconcileInFlight failed');
  });
} catch (err) {
  this.logger.warn({ err }, 'boot reconcileInFlight unavailable');
}
```

A boot sweep is a nice-to-have that buys one cycle of latency on a repair path where nobody is waiting. It should never be able to prevent the dispatcher from starting.

christrudelpw (commented):
Review of #1076 (issue #1058) — **3 findings, 0 blocking.** All inline.

Q1=D landed as asked, and it landed for the right reason. The two-sweep tracker is what makes this safe, and the mechanism is correct: candidates are keyed by itemKey with the sweep-id of first sighting, fire only on a strictly-earlier sighting, and the self-clear loop drops any tracker entry that is no longer residue — so a transient artifact of the non-atomic snapshot vanishes on its own and never reaches an `SREM`.

I spent most of this review trying to break that, and want to record what I checked so the next reader doesn't have to redo it:

- **The snapshot race I raised in Q1 is real and is closed.** Reads go in-flight → pending → claimed, so the vulnerable transition is `claimed → pending` (the `release()` retry path), where an item can appear in neither snapshot. Two-sweep closes it because the item is back in pending 30s later. The tracker self-clear is what makes "consecutive" actually mean consecutive.
- **The boot-sweep/reaper interleaving I worried about is closed too.** If the boot sweep and the first reaper sweep could overlap, both could observe the *same* transient window and the two sightings would not be a cycle apart — defeating the whole guarantee. `reaperLoop` sleeps `heartbeatCheckIntervalMs` at the top of its loop before the first cycle, so the fire-and-forget boot sweep completes long before. Q2=B's "arm at t=0, repair at t=30s" works as intended.
- **Nothing can re-add a residue member mid-sweep.** Since #1060, `enqueue()` and `enqueueIfAbsent()` share one in-flight-guarded Lua, and I found no other `SADD` against `IN_FLIGHT_KEY`. This is the invariant the whole design rests on — see finding 1, because it is currently undocumented while a comment that *is* documented says something false about the same window.
- **Q3=C, Q4=B, Q5** all honoured: both caches cleared on `SREM`, per-cycle cap with an aggregate line and sampled keys, and no scratch key at all (so the CROSSSLOT hazard is moot — `numberOfKeys: 1`, stated explicitly).

The findings are, in order:

1. **The script's docblock describes a guard it doesn't implement** — a concurrent re-add yields `SISMEMBER` = 1, so the documented "either way, we do NOT SREM" is inverted, and `SISMEMBER`+`SREM` is behaviourally a bare `SREM`. Harmless today only because of the invariant above, which is the thing that should be written there instead. `skippedRaceReappeared` also counts the inverse of its name.
2. **Boot sweep isn't guarded against a synchronous throw** — `.catch()` doesn't catch it, and in `start()` the blast radius is the whole dispatcher. This is #1054 finding 5's hazard, quoted correctly 25 lines below and not applied 300 lines above. Not reachable today (both adapters implement the method); one line to keep it that way.
3. **`ageMs` is structurally null for cross-process residue** — the exact population the boot sweep exists for. Not recoverable; just shouldn't imply otherwise.

None of these touch the correctness of the sweep. Take them in one round and I'll merge.

packages/orchestrator/src/services/redis-queue-adapter.ts
```diff
@@ -752,6 +844,181 @@ export class RedisQueueAdapter implements QueueManager {
...
+          { err: error, event: 'orphan-in-flight-lua-error', itemKey },
+          'Redis error in RECONCILE_IN_FLIGHT_SCRIPT, retaining tracker entry for next cycle',
+        );
+        continue;
+      }
+
+      if (result === 1) {
+        // AD-6: full cache cleanup matches complete() / dead-letter / reclaim.
+        const enqueuedAtMs = this.enqueuedAtCache.get(itemKey);
+        const ageMs = enqueuedAtMs !== undefined ? now - enqueuedAtMs : null;
```
christrudelpw, line 1000: `ageMs` is sourced only from the process-local `enqueuedAtCache`, so it is structurally `null` for residue that survived a process restart — which is the primary scenario the boot sweep exists to serve.

**Failure scenario:** `this.enqueuedAtCache` is an in-process `Map`, empty at boot. Residue that outlived the previous process therefore always resolves to `enqueuedAtMs === undefined` → `ageMs: null` on the `orphan-in-flight-reconciled` warn.

That is the case this feature is most for. My own Q2=B answer argued the boot sweep matters because *"the likeliest source of real residue is whatever was in flight when the previous process died"* — and for exactly that population, the age field is guaranteed empty. An operator investigating a post-crash repair sees `ageMs: null` and cannot tell whether it means "unknown" or "the lookup is broken".

The age is not recoverable — `IN_FLIGHT_KEY` is a plain set of itemKeys with no timestamps, and a residue member has neither a pending payload nor a claim hash to read `enqueuedAt` from. `hasInFlightAge()` would not help for the same reason. So this is not a request to find the value; it is a request to stop implying it should be there.

**Fix — either is fine:**
- Omit the key when null, matching the presence-vs-value discipline used elsewhere in this batch (`...(ageMs !== null ? { ageMs } : {})`), so its absence reads as "not applicable" rather than "lookup failed"; or
- Keep it and add one line to the docstring: *`ageMs` is null for residue predating this process — the cache is process-local and the set carries no timestamps. Expected on boot-sweep repairs.*

Lowest-priority of the three; I'd take it in the same round as the others rather than separately.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in d327e89f — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/worker-dispatcher.ts
christrudelpw: Addressed in d327e89f — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/services/redis-queue-adapter.ts
christrudelpw: Addressed in d327e89f — please review, and re-open this thread if it still falls short.

PR #1078, opened 2026-07-29
feat: #1067 1067-problem-generacy-1053-s

christrudelpw:
## Summary

Closes #1067

---


packages/orchestrator/src/routes/cockpit-gates.ts
```diff
@@ -59,6 +59,14 @@ const GateQueryStringSchema = z
...
     gateType: GateTypeSchema.optional(),
     generation: z.string().min(1).optional(),
+    /**
+     * #1067 — Passive pass-through. Route MUST NOT enforce
+     * "runId ⇒ generation" here — the cloud route enforces it authoritatively
+     * (`services/api/src/routes/clusters/cockpit-gates.ts` .refine). Duplicating
+     * the refinement would mask the cloud's RFC-7807 400 behind a route-side
+     * 'VALIDATION' 400.
+     */
+    runId: z.string().min(1).optional(),
```
christrudelpw, line 72: In list mode this route accepts `runId`, never forwards it, and never rejects it — a silent 200 over an unfiltered list. The docblock says the cloud enforces the constraint authoritatively, but in list mode the cloud is never sent the parameter, so nothing enforces it anywhere.

**Failure scenario:** Trace a list-mode request carrying `runId` through this route:

- `:239` destructures `runId` from `parsed.data` — the schema accepts it.
- `:240` sets `mode = generation !== undefined ? 'status' : 'list'`, so list mode means `generation === undefined`.
- `:248` forwards `runId` on the **status** branch only.
- `:264` is the list branch: `client.listGates({ issueRef, ...(gateType ? { gateType } : {}) })` — no `runId`.

So `GET /cockpit/gates?issueRef=owner%2Frepo%23123&runId=run-abc` returns `200` with every run's gates. The caller asked a run-scoped question and got an unscoped answer that looks successful. That is the same silent-swallow I asked generacy-cloud#893 to close, reappearing one layer up — in the PR that cites the resulting refine as its justification.

**The docblock's reasoning is inverted for the only case it governs:**

> *Route MUST NOT enforce "runId ⇒ generation" here — the cloud route enforces it authoritatively … Duplicating the refinement would mask the cloud's RFC-7807 400 behind a route-side 'VALIDATION' 400.*

Work through when a route-side refine could actually fire. It fires only when `runId !== undefined && generation === undefined` — which is exactly list mode. And in list mode the cloud never receives `runId`, so there is no RFC-7807 400 to mask. In status mode `generation` is present by construction, so the refine can never fire there and cannot mask anything either.

The refine therefore masks **nothing, in either mode**, and the stated reason for omitting it does not describe a reachable situation. Meanwhile the comment tells the next person auditing "is `runId` validated on this surface?" that the cloud has it covered — and they will stop reading. That is a false safety claim of the same kind as the two corrected in tonight's other reviews (`console.debug`'s severity note on generacy-cloud#893, and #1076's script docblock describing a guard it did not implement). Both were mechanically harmless; both would have misled the next reader into building on a false premise.

**Fix** — add the sibling refine, mirroring the cloud and the `generation ⇒ gateType` rule already directly below it:

```ts
.refine((v) => v.runId === undefined || v.generation !== undefined, {
  message: 'runId requires generation',
  path: ['runId'],
})
```

and replace the docblock with what is actually true: the refine is safe to duplicate precisely *because* it can only fire in list mode, where the cloud is never consulted; status-mode requests still reach the cloud and still surface its RFC-7807 body unchanged.

The alternative — forwarding `runId` on the list branch so the cloud does the rejecting — is worse. It converts a local 400 into a network round trip whose only possible outcome is a 400, and it contradicts Q1=C, which exists because forwarding `runId` on list breaks the sweep's primary dedup primitive.

Non-blocking: no caller sends `runId` to this route yet (Phase C is agency#469), so nothing is broken today. That is also why it is cheap to close now, before the first caller is written against the looser shape.

christrudelpw (commented):
Review of #1078 (issue #1067, Phase B) — **1 finding, 0 blocking.** Inline.

All five clarification answers landed as specified, including the two that were forced by Phase A's deployed contract rather than chosen:

- **Q1=C** — `runId` is accepted on `CockpitGateListInputSchema` for surface parity and dropped in the handler before `listGates`, with a comment at the drop site quoting the cloud refine that requires it. Exactly the shape asked for, and the reason is recorded where someone would otherwise "fix" it by forwarding.
- **Q2=A** — `url.searchParams.set('runId', …)`, camelCase, with the cloud commit pinned in the comment.
- **Q3=C** — only `cockpit_gate_status` logs, post-call, with B's richer field set, **and on the failure path** with `resolvedStatus: 'error'` plus the error text. The addition I did not ask for and should have: the log records `runIdSource` as `'explicit' | 'unset'` and never the value, because auto-run ids embed cluster/repo/issue/timestamp. That is a better call than what I specified.
- **Q4=A** — three-tool identity matrix in `parity-gate-tuple-identity.test.ts`, both tuple shapes.
- **Q5=C** — both halves present.

Things I checked rather than assumed, recorded so the next reader doesn't repeat them:

- **The MCP inputSchema landmine is avoided.** Both `CockpitGateStatusInputSchema` and `CockpitGateListInputSchema` stay flat `z.object(…).strict()`. Had either been widened via intersection, `.shape` would be undefined and the tool's advertised inputSchema would silently go empty — generacy#1032/#1033, which cost a dogfood round.
- **`cloud-gate-query-client.ts:357` passing `runId: input.runId` unconditionally is safe**, despite breaking the spread-guard idiom used everywhere else in this PR. `buildUrl` filters with `if (v !== undefined) base.searchParams.set(k, v)`, and its parameter type is `Record<string, string | undefined>`, so undefined is contract, not accident. I started writing this up as a finding — `runId=undefined` as a literal query string would compose a 4-segment key ending in the string "undefined" and turn every status lookup into a silent `absent` — then read the helper. Worth knowing that the filter in `buildUrl` is the only thing standing between here and that bug.
- **FR-005 byte-identity is genuinely pinned**, and I nearly filed a wrong finding claiming otherwise. I grepped the MCP `__tests__` directory, found only `toContain` assertions, and started arguing the snapshot half of Q5 was missing. It is in the orchestrator's tree: `cloud-gate-query-client.runid.test.ts:123` asserts `expect(actualUrl).toBe(URL_STATUS_NO_RUNID)` against a checked-in constant, with `:166` covering the with-`runId` case and `:145`/`:194-195` supplying the structural half. That combination catches the added-parameter regression a `toContain` suite cannot.
- **The new `getLogger` import does not violate observer independence.** The scan forbids `../gates/client.js`, `./cockpit_gate_open.js`, `./cockpit_gate_ack.js`, and any path containing `retain`; `utils/logger.js` matches none.

The one finding is in the **orchestrator's** `GET /cockpit/gates` route rather than the MCP tools: in list mode it accepts `runId`, forwards it nowhere, and rejects it never. The docblock explains the missing refine by saying the cloud enforces the rule authoritatively — but in list mode the cloud is never sent the parameter, so nothing enforces it at any layer. One refine closes it.

packages/orchestrator/src/routes/cockpit-gates.ts
christrudelpw: Addressed in 2b359897 — please review, and re-open this thread if it still falls short.

PR #1079, opened 2026-07-29
feat: #1077 1077-problem-after-generacy-1066

christrudelpw:
## Summary

Closes #1077

---

### DIFF
--- a/packages/cluster-relay/src/relay.ts
+++ b/packages/cluster-relay/src/relay.ts
@@ -20,6 +20,23 @@ export interface Logger {
   error(obj: Record<string, unknown>, msg: string): void;
 }
 
+/**
+ * Metadata for a pending outbound cockpit frame (#1077). Registered by the
+ * orchestrator's route immediately after mint via `registerPendingFrame()`;
+ * carried onto the settle log when the cloud replies.
+ */
+export interface PendingFrameMeta {
+  frameType: 'gate-open' | 'gate-outcome';
+  gateId: string;
+}
+
+interface PendingFrame {
+  frameType: string;
+  gateId: string;
+  registeredAt: number;
+  ttlHandle: ReturnType<typeof setTimeout>;
+}
+
 /** Options accepted by ClusterRelayClient (orchestrator-facing API). */
 export interface ClusterRelayClientOptions {
   /** API key for cloud authentication (GENERACY_API_KEY) */
@@ -75,6 +92,8 @@ export class ClusterRelay {
   private heartbeatTimer: ReturnType<typeof setInterval> | null = null;
   private pongReceived = true;
   private metadataOverride: Partial<ClusterMetadata> | null = null;
+  private readonly pendingFrames = new Map<string, PendingFrame>();
+  private static readonly TTL_MS = 30_000;
 
   /**
    * Accept either a full RelayConfig or a ClusterRelayClientOptions (orchestrator API).
@@ -189,7 +208,15 @@ export class ClusterRelay {
    * Gracefully disconnect.
    */
   async disconnect(): Promise<void> {
-    if (!this.running) return;
+    if (!this.running) {
+      // Even when the outer loop is not running, a leftover pending-frame map
+      // could exist if register was called pre-connect. Clear defensively.
+      for (const entry of this.pendingFrames.values()) {
+        clearTimeout(entry.ttlHandle);
+      }
+      this.pendingFrames.clear();
+      return;
+    }
 
     this._state = 'disconnecting';
     this.running = false;
@@ -211,6 +238,11 @@ export class ClusterRelay {
       });
     }
 
+    for (const entry of this.pendingFrames.values()) {
+      clearTimeout(entry.ttlHandle);
+    }
+    this.pendingFrames.clear();
+
     this._state = 'disconnected';
   }
 
@@ -223,6 +255,38 @@ export class ClusterRelay {
       return;
     }
     this.ws.send(JSON.stringify(message));
+  }
+
+  registerPendingFrame(frameId: string, meta: PendingFrameMeta): void {
+    if (!frameId) {
+      this.logger.debug({ frameId }, 'registerPendingFrame ignored empty frameId');
+      return;
+    }
+    const existing = this.pendingFrames.get(frameId);
+    if (existing) {
+      clearTimeout(existing.ttlHandle);
+    }
+    this.pendingFrames.set(frameId, {
+      frameType: meta.frameType,
+      gateId: meta.gateId,
+      registeredAt: Date.now(),
+      ttlHandle: setTimeout(() => this.evictOnTtl(frameId), ClusterRelay.TTL_MS),
+    });
+  }
+
+  private evictOnTtl(frameId: string): void {
+    const entry = this.pendingFrames.get(frameId);
+    if (!entry) return;
+    this.pendingFrames.delete(frameId);
+    this.logger.debug(
+      { frameId, ageMs: Date.now() - entry.registeredAt },
+      'cluster.cockpit pending frame evicted on TTL',
+    );
+  }
+
+  /** @internal Test-only accessor for the pending-frame map size. */
+  _pendingFramesSizeForTests(): number {
+    return this.pendingFrames.size;
   }
 
   /**
@@ -327,22 +391,39 @@ export class ClusterRelay {
           return;
         }
 
-        // Cloud-sent gate acknowledgement (#1063). Observability-only until
-        // #1059 steps 4-7 wire the frameId correlation. The return is
-        // structural (Q3=A / FR-003) — prevents any future subscriber from
-        // beginning to correlate before the full design lands.
+        // Cloud-sent gate acknowledgement (#1063 + #1077). #1077 wires the
+        // frameId-keyed pending map: settle on match, quiet-drop on miss.
+        // The return is structural (Q3=A / FR-003) — registered onMessage
+        // handlers must not observe cluster.cockpit.reply.
         if (message.type === 'cluster.cockpit.reply') {
-          if (message.accepted) {
-            this.logger.debug({ message }, 'cluster.cockpit.reply received');
+          const { frameId } = message;
+          if (frameId !== null && this.pendingFrames.has(frameId)) {
+            const entry = this.pendingFrames.get(frameId)!;
+            clearTimeout(entry.ttlHandle);
+            this.pendingFrames.delete(frameId);
+            this.logger.info(
+              {
+                frameId,
+                frameType: message.frameType,
+                gateId: message.gateId,
+                accepted: message.accepted,
+                reason: message.reason,
+                priorStatus: message.priorStatus,
+                ageMs: Date.now() - entry.registeredAt,
+              },
+              'cluster.cockpit.reply settled pending frame',
+            );
           } else {
             this.logger.info(
               {
-                reason: message.reason,
+                frameId,
                 frameType: message.frameType,
                 gateId: message.gateId,
+                accepted: message.accepted,
+                reason: message.reason,
                 priorStatus: message.priorStatus,
               },
-              'cluster.cockpit.reply dropped',
+              'cluster.cockpit.reply had no matching pending frame',
             );
           }
           return;
--- a/packages/cluster-relay/src/relay.ts
+++ b/packages/cluster-relay/src/relay.ts
@@ -278,8 +278,16 @@ export class ClusterRelay {
     const entry = this.pendingFrames.get(frameId);
     if (!entry) return;
     this.pendingFrames.delete(frameId);
+    // Carry the stored frameType/gateId onto the eviction log — this is the
+    // one path where the reply never arrives, so the entry's own copy is the
+    // only source available for joining the evicted frameId back to the gate.
     this.logger.debug(
-      { frameId, ageMs: Date.now() - entry.registeredAt },
+      {
+        frameId,
+        frameType: entry.frameType,
+        gateId: entry.gateId,
+        ageMs: Date.now() - entry.registeredAt,
+      },
       'cluster.cockpit pending frame evicted on TTL',
     );
   }
@@ -401,6 +409,22 @@ export class ClusterRelay {
             const entry = this.pendingFrames.get(frameId)!;
             clearTimeout(entry.ttlHandle);
             this.pendingFrames.delete(frameId);
+            // Integrity cross-check: the cloud echoes the reply off the raw
+            // frame payload, so a disagreement between the entry we stored at
+            // registerPendingFrame time and the entry the cloud derived from
+            // the same frame implies a frameId collision or a cloud-side mixup.
+            if (entry.gateId !== message.gateId) {
+              this.logger.warn(
+                {
+                  frameId,
+                  storedGateId: entry.gateId,
+                  replyGateId: message.gateId,
+                  storedFrameType: entry.frameType,
+                  replyFrameType: message.frameType,
+                },
+                'cluster.cockpit.reply gateId disagrees with pending entry',
+              );
+            }
             this.logger.info(
               {
                 frameId,
--- a/packages/orchestrator/src/routes/cockpit-gates.ts
+++ b/packages/orchestrator/src/routes/cockpit-gates.ts
@@ -1,3 +1,4 @@
+import { randomBytes } from 'node:crypto';
 import { z, ZodError } from 'zod';
 import type { FastifyInstance, FastifyRequest, FastifyReply } from 'fastify';
 // Canonical frozen wire schemas (tetrad-development/docs/cockpit-remote-gates-plan.md
@@ -86,6 +87,10 @@ const GateQueryStringSchema = z
  * negatives collapse to `absent`.
  */
 type ThreeState = 'open' | 'answered' | 'absent';
+
+function mintFrameId(): string {
+  return `frm_${randomBytes(12).toString('hex')}`;
+}
 
 function collapseCloudStatus(status: CloudGateStatusResponse['status']): ThreeState {
   switch (status) {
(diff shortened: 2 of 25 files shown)


christrudelpw (commented):
**Changes requested** — two findings, both inline, both about the pending map rather than the mint.

The mint half is right and I checked the parts that were most likely to be wrong. `frm_` + 12 random bytes matches the fixture convention #1066 established; minting before `tryEmitOrRetain` means a retained frame carries its id into the queue and the 202 can return it; `parsed.frameId ?? mintFrameId()` honours a caller-supplied id, which is exactly what #1066's `''`/`null` → `undefined` transforms exist to make safe. Settle evicts as well as TTL. The `cluster.cockpit.reply` handler still returns structurally so no `onMessage` subscriber can observe it.

I also verified the loop actually closes, since that was the whole risk with this issue: the cloud reads `frameId` off the raw frame data and echoes it on every reply, including the unknown-subtype branch (`generacy-cloud/services/api/src/services/relay/message-handler.ts:812-849`). And `getRelayClient()` resolves in the same process that owns the relay client, so the registering side and the settling side share memory — the architectural point that made Q2's option A impossible.

Local verification on the branch: `cluster-relay` + `orchestrator` build clean, `relay.test.ts` 29/29, `cockpit-gates.test.ts` + `retained-cockpit-events.test.ts` 57/57, and `cockpit-gates-frameid.integration.test.ts` 5/5 under `vitest.integration.config.ts` (it is excluded from the default run by design and picked up by CI's separate `integration` job — worth knowing before reading a green `pnpm test` as covering it).

The two findings:

1. **`cockpit-gates.ts:342`** — the TTL is armed at accept, not at send, so a retained frame's entry expires during the outage that caused the retention. Reconnect backoff reaches 35s cumulative by the third retry against a 30s TTL. `relay.test.ts:1032` pins the timer as *preserved* across reconnect, which is the behaviour that produces this. Fix is a re-arm in `drainInto`, plus the retain → drain → echo → settle test that is currently missing.
2. **`relay.ts:283`** — `frameType`/`gateId` are stored on every pending entry and read nowhere. The eviction log, which is the only place the stored copy is the sole source, omits them, so an eviction cannot be joined back to a gate.

Neither is a correctness bug — in both cases the degraded outcome is the pre-#1077 log line. But finding 1 removes the retain-path correlation that #1066's preserve-across-retain work exists to enable, and finding 2 blunts the eviction log that finding 1 would otherwise be diagnosed from. They compound, which is why I would rather fix both now than discover them together from a quiet log later.

packages/orchestrator/src/routes/cockpit-gates.ts
```diff
@@ -332,26 +337,33 @@ export function setupCockpitGatesRoute(
     async (request: FastifyRequest, reply: FastifyReply) => {
       try {
         const parsed = GateOpenSchema.parse(request.body);
+        const frameId = parsed.frameId ?? mintFrameId();
+        const emitData = { ...parsed, frameId };
+        options.getRelayClient()?.registerPendingFrame(frameId, {
```
christrudelpw, line 342: The 30s pending-frame TTL is armed at request-accept, but the reply can only arrive after the send. For a retained frame those are separated by the whole outage, so retained frames' echoes land in the quiet-drop branch — the exact outcome the 30s TTL was chosen to avoid.

**Failure scenario:** `registerPendingFrame` is called at request-accept time, immediately before `tryEmitOrRetain`. That is the right place to *mint* — the 202 needs the id, and a retained frame must carry its id into the queue. But it starts the 30-second TTL clock at **accept**, and the reply can only arrive after the **send**. For a frame that is retained, those two moments are separated by the entire outage.

**The retain path is exactly the path that cannot survive this.** A frame is retained precisely because the relay is not connected, and the retain queue drains on reconnect (`relay-bridge.ts:212` → `drainRetainedCockpitEvents` → `retainer.drainInto(client)`). Reconnect backoff is exponential from a 5s base (`relay.ts:526`, `baseReconnectDelayMs * 2^attempt`), so cumulative wait runs 5s → 15s → 35s → 75s. **From the third retry onward the pending entry has already been evicted before the frame is put on the wire.** Its echo then lands in the `'cluster.cockpit.reply had no matching pending frame'` branch — indistinguishable from a reply for a frame this process never sent.

There is a second, simpler route to the same end state: `options.getRelayClient()?.registerPendingFrame(...)` is optional-chained, so when `getRelayClient()` returns `null` **no entry is registered at all** — yet the handler still returns `202 {accepted: true, retained: true, frameId}`. The response names a correlation id that has no pending entry and never will. `tryEmitOrRetain` treats `client === null` as a retain case (`:184`), so this is a live path, not a theoretical one.

`drainInto` sends `head.data` verbatim, which is correct and is what #1066 Q3's preserve-across-retain was for — the `frameId` survives the queue. What does not survive is the entry that would match it.

**This specifically undoes the reason 30s was chosen over 5s.** From the Q3 answer: *"5s is too short: it does not cover a relay reconnect, and #1066 Q3 deliberately preserves `frameId` across retain-and-drain precisely so a post-reconnect drain still correlates. A TTL shorter than the reconnect window would guarantee those replies land in the quiet-drop branch, making the retain-path correlation work pointless."* Anchoring the clock at accept reintroduces that outcome at 30s. The TTL is not too short; it is being measured from the wrong event.

I want to be clear this is not a correctness bug — nothing is delivered twice, nothing is lost, and the degraded state is the pre-#1077 log line. It is that the retain path, which is the one case where correlation has real diagnostic value (did the frames I queued during the outage actually land?), is the one case that structurally cannot correlate.

**Fix — re-arm at emit, not only at accept.** Keep the mint and the accept-time `registerPendingFrame` exactly as they are (they cover the connected case and give the 202 its id). Add a re-arm in `drainInto`, which already holds both the client and the payload:

```ts
drainInto(client) {
  ...
  const d = head.data as { frameId?: unknown; type?: unknown; gateId?: unknown };
  if (typeof d.frameId === 'string' && d.frameId) {
    // Re-arm the TTL at send time — the accept-time entry may have expired
    // during the outage that caused the retain (#1077).
    client.registerPendingFrame(d.frameId, {
      frameType: d.type as PendingFrameMeta['frameType'],
      gateId: String(d.gateId ?? ''),
    });
  }
  client.send({ type: 'event', event: 'cluster.cockpit', data: head.data, timestamp: head.timestamp });
  ...
}
```

`registerPendingFrame` already handles re-registration (`clearTimeout(existing.ttlHandle)` at `:266-268`), so this is idempotent by construction and needs no new branch. It also closes the `getRelayClient() === null` hole, since drain always has a client.

**Please add the test that is missing.** The relay suite covers settle, quiet-drop, TTL eviction, `disconnect()` clearing, and — pointedly — *"transient reconnect (ws close) preserves the pending map and the TTL timer"* (`relay.test.ts:1032`). That last one shows the reconnect case was considered, and it pins the timer as **preserved**, which is precisely the behaviour that expires the entry mid-outage. What no test covers is **retain → reconnect → drain → echo → settle**. Add it with fake timers advanced past 30s before the drain; against today's code it fails, which is the point.

If instead the intended semantics really are "the TTL is a wall-clock bound from acceptance, and retained frames are not expected to correlate," then say so in `contracts/pending-map.md` and mark `relay.test.ts:1032` as pinning that decision — but that reverses the Q3 rationale and should be an explicit choice, not a side effect of where the call sits.

packages/cluster-relay/src/relay.ts
```diff
@@ -225,6 +257,38 @@ export class ClusterRelay {
...
+    });
+  }
+
+  private evictOnTtl(frameId: string): void {
+    const entry = this.pendingFrames.get(frameId);
+    if (!entry) return;
+    this.pendingFrames.delete(frameId);
+    this.logger.debug(
+      { frameId, ageMs: Date.now() - entry.registeredAt },
+      'cluster.cockpit pending frame evicted on TTL',
```
christrudelpw, line 291: `PendingFrameMeta.frameType` and `.gateId` are stored on every pending entry and never read. The TTL-eviction log — the one place the stored copy is the only available source — omits them, leaving evictions un-actionable.

**Failure scenario:** `PendingFrameMeta` carries `frameType` and `gateId`, `registerPendingFrame` stores both on the `PendingFrame` record, and **nothing ever reads either one.** The only fields consumed anywhere are `ttlHandle` (`:215`, `:242`, `:267`, `:402`) and `registeredAt` (`:282`, `:412`).

On settle, every field on the log line except `ageMs` comes from the inbound reply, not from the entry:

```ts
const entry = this.pendingFrames.get(frameId)!;
...
frameType: message.frameType,   // reply
gateId: message.gateId,         // reply
ageMs: Date.now() - entry.registeredAt,
```

And the eviction log — the one place the stored copy is the *only* source available — omits them:

```ts
this.logger.debug(
  { frameId, ageMs: Date.now() - entry.registeredAt },
  'cluster.cockpit pending frame evicted on TTL',
);
```

**That is the line that most needs them.** An eviction means a frame was sent and no reply ever came back for it, which is the failure this whole map exists to make visible. As written, the operator gets `frm_9f3c…` and an age. To learn which gate silently failed, they have to grep backwards through the log for the mint that produced that id — and the mint is logged in a *different package* (`cockpit-gates.ts`'s `'cockpit gate emitted'` line carries `gateId` and `type` but not `frameId`, so even that grep does not join). The eviction is therefore un-actionable in practice.

This defeats the specific job the eviction log was asked to do. From the Q3 answer: *"Please log evictions at `debug` with the `frameId` and the age. A TTL that is too short is otherwise invisible... An eviction log turns 'is 30s enough?' into a question the data answers."* It answers *how often* — but not *for which frames*, which is what tells you whether the evictions are the retain path (finding 1) or something else.

**Fix, one line:**

```ts
this.logger.debug(
  {
    frameId,
    frameType: entry.frameType,
    gateId: entry.gateId,
    ageMs: Date.now() - entry.registeredAt,
  },
  'cluster.cockpit pending frame evicted on TTL',
);
```

**Second use worth taking while the data is there.** On settle, `entry.gateId` and `message.gateId` should agree — the cloud echoes `frameId` off the raw frame and derives its reply from the same payload (`generacy-cloud message-handler.ts:812-825`). A disagreement means a `frameId` collision or a cloud-side mix-up, and today it would pass through silently because the log takes the cloud's word for it. Logging `entry.gateId` alongside `message.gateId`, or warning when they differ, turns the stored copy into an integrity check instead of dead weight. Cheap, and it is the only cross-check available on this path.

If neither use is wanted, the honest alternative is to drop `frameType`/`gateId` from `PendingFrameMeta` and `PendingFrame` entirely and let `registerPendingFrame(frameId)` take one argument — but that also removes the interface change from `types/relay.ts` and the `fake-peer.ts` plumbing, so keeping the fields and using them is clearly the better trade.

packages/orchestrator/src/routes/cockpit-gates.ts
christrudelpw: Addressed in 22bbdaa8 — please review, and re-open this thread if it still falls short.

packages/cluster-relay/src/relay.ts
christrudelpw: Addressed in 22bbdaa8 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/routes/cockpit-gates.ts
```diff
@@ -332,26 +337,33 @@ export function setupCockpitGatesRoute(
     async (request: FastifyRequest, reply: FastifyReply) => {
       try {
         const parsed = GateOpenSchema.parse(request.body);
+        const frameId = parsed.frameId ?? mintFrameId();
```
christrudelpw, line 340: **BLOCKING** Per-emit frameId minting breaks the #1053 FR-008 integration test's byte-identity assertion on a re-POSTed body (CI red). The behaviour is correct — per-request-body minting would corrupt the pending map — so the assertion must exclude `frameId` and pin that it differs.

**Failure scenario:** **CI is red on this branch.** `packages/orchestrator/src/__tests__/cockpit-gates-integration.integration.test.ts:351` fails in the `integration` job (run 30463690925, `1 failed | 111 passed`):

```
FAIL  #1053 FR-008 — Run A `applied` then Run B re-emits same natural gate;
      distinct gateIds at peer, within-run frames byte-identical

AssertionError: expected { type: 'gate-open', …(14) } to deeply equal { … }
-   "frameId": "frm_000ee66695e81b35ec92bfd2",
+   "frameId": "frm_5516f113b9563c0600fea544",
```

Every other field matches — same `gateId`, same `gateKey`, same `askedAt`. Only the minted `frameId` differs.

This is not in the PR's changed files, which is why the suites run for this feature were all green: the regression is in a **#1053** test that this feature invalidates.

**The behaviour is right; the assertion is now over-broad.** The test POSTs the *same request body* twice — *"mimicking a MCP-tool retry after the askedAt hoist"* — and asserts `expect(runAFrames[0]!.data).toEqual(runAFrames[1]!.data)`. That held when frames carried no per-frame identity. `frameId` is by definition per-**frame**, and two POSTs are two frames, so the two ids must differ.

**Minting per request body instead would break the pending map.** If both POSTs shared a `frameId`, the second `registerPendingFrame` would clear the first's TTL handle and overwrite its entry (`relay.ts:265-268`), so the two replies the cloud sends would settle against one entry — the first reply settles it, the second finds nothing and lands in `'had no matching pending frame'`. Per-emit minting is what makes the map correct, and it is what Q1 asked for (*"per-frame is most faithfully per emit"*).

**SC-002's actual invariant survives.** Its stated purpose is *"Cloud dedup collapses these to one inbox row"* — and cloud dedup for `gate-open` is an upsert on the identity-derived doc path `organizations/{orgId}/cockpitGates/{gateId}`, which is unaffected by `frameId`. One inbox row still results. What changed is only the literal byte-identity of the two wire frames.

**Fix — re-pin, do not weaken.** Do not delete the `toEqual`. Exclude the field that is *designed* to differ and assert that it differs, which makes the test stronger than it was:

```ts
// Byte-identity: both frames carry the same gateId + gateKey + askedAt.
// `frameId` is EXCLUDED — #1077 mints one per emit, so two POSTs of the same
// body are two frames with two ids by design. Asserting they DIFFER pins that
// property here rather than losing it.
const { frameId: frameIdA1, ...restA1 } = runAFrames[0]!.data as Record<string, unknown>;
const { frameId: frameIdA2, ...restA2 } = runAFrames[1]!.data as Record<string, unknown>;
expect(restA1).toEqual(restA2);
expect(frameIdA1).toEqual(expect.stringMatching(/^frm_[0-9a-f]{24}$/));
expect(frameIdA2).toEqual(expect.stringMatching(/^frm_[0-9a-f]{24}$/));
expect(frameIdA1).not.toBe(frameIdA2);
```

Also update the test's SC-002 comment (`:334-337`), which currently says the retry *"reaches the peer as a byte-identical second event"* — after this PR it reaches the peer as an identical-except-`frameId` second event, and the dedup guarantee is the `gateId` upsert, not byte-identity.

**One process note worth acting on.** I verified this branch by running only the suites the PR touches, and that was not sufficient — the break is in a file the PR does not touch. The `integration` job is the only thing that caught it, and it is a separate CI job from `ci`, gated on `draft == false`, so it produced no signal at all while the PR sat in draft. Worth running `pnpm -r --if-present run test:integration` locally before the next `gh pr ready` on a cockpit-gates change; I will.

christrudelpw (commented):
**Second round — one blocking, and it is a CI failure, not a review opinion.**

Both first-round findings are fixed well, and the optional extra was taken:

- `drainInto` now re-arms the pending-frame TTL at send time, with a type guard, and a comment that spells out the 5s→15s→35s→75s backoff maths. It also closes the `getRelayClient() === null` accept-time path where no entry was registered at all.
- The eviction log carries `frameType` + `gateId`.
- The settle path now warns when `entry.gateId !== message.gateId` — the integrity cross-check I raised as optional. Good call to take it; it is the only place a `frameId` collision or cloud-side mixup could surface.
- `retain → reconnect → drain → echo settles the pending frame (#1077 review)` exists in the integration suite, which is exactly the missing case.

Local verification on the fixed branch: builds clean, `relay.test.ts` 31/31, orchestrator unit 60/60, `cockpit-gates-frameid.integration.test.ts` 6/6.

**What that verification missed, and CI did not:** `cockpit-gates-integration.integration.test.ts:351` — the #1053 FR-008 test — now fails, because it POSTs the same body twice and asserts the two peer frames are byte-identical. Per-emit `frameId` minting makes that false by design. The behaviour is right (minting per request body instead would let the second `registerPendingFrame` clobber the first's entry, so one of the two replies would be miscorrelated), and SC-002's real guarantee — one inbox row, via the `gateId` upsert — is unaffected. Only the literal byte-identity assertion needs re-pinning: exclude `frameId` from the `toEqual` and assert that the two ids differ, which makes the test stronger rather than weaker. Exact replacement in the inline comment, along with the stale SC-002 prose above it.

The `integration` job is separate from `ci` and gated on `draft == false`, so it gave no signal while this sat in draft — worth running `test:integration` locally before the next `gh pr ready` on anything touching the cockpit-gates route.

packages/orchestrator/src/routes/cockpit-gates.ts
christrudelpw: Addressed in cdd94661 — please review, and re-open this thread if it still falls short.

christrudelpw (commented):
**Approved** (posted as COMMENT — GitHub rejects self-approval). Both rounds resolved, CI green (ci + integration).

Round 2 re-pinned the #1053 FR-008 assertion the way I asked: excludes `frameId` from the deep-equal, asserts both ids match `/^frm_[0-9a-f]{24}$/`, and asserts they differ — so the per-emit property is pinned rather than lost. SC-002's comment now names the `gateId` upsert as the dedup guarantee instead of byte-identity, which is the accurate statement.

Verified locally on the branch: builds clean, `relay.test.ts` 31/31, orchestrator unit 60/60, both cockpit-gates integration suites 21/21. The four redis-queue-adapter integration failures I hit locally reproduce on `develop` (no local Redis) and are unrelated.

The correlation loop is closed end to end: the route mints `frm_` + 12 random bytes per emit, registers a pending entry, `drainInto` re-arms the TTL at send time so a retained frame survives the outage that retained it, the cloud echoes `frameId` off the raw frame on every reply path, and the relay settles or quiet-drops with the gate named in both the settle and eviction logs.

PR #1081, opened 2026-07-29
feat: #1068 1068-problem-gate-identity-work

christrudelpw:
## Summary

Closes #1068

---


packages/orchestrator/src/__tests__/cockpit-gates/no-simulate-phase-in-src.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      '  packages/generacy/src/cli/commands/cockpit',
+      '  packages/cockpit/src',
+      '  --exclude-dir=__tests__',
+      '  --exclude-dir=tests',
+      "  --exclude='*.test.ts'",
+      "  --exclude='*.spec.ts'",
+      '  || true',
+    ].join(' ');
+    const output = execSync(cmd, {
+      cwd: '/workspaces/generacy',
```
christrudelpw: **BLOCKING** The FR-012 boundary test hardcodes `cwd: '/workspaces/generacy'`, which does not exist on a CI runner — `execSync` throws ENOENT and the `ci` job goes red. Deleting the option instead makes the test pass while grepping nothing.

**Failure scenario:** `cwd: '/workspaces/generacy'` is this devcontainer's path. On a GitHub runner the repo is at `/home/runner/work/generacy/generacy`, so that directory does not exist.

`execSync` throws when `cwd` is missing — I verified rather than assumed:

```
$ node -e "require('node:child_process').execSync('grep -rE X . || true', {cwd:'/nonexistent-repo-path'})"
THREW: ENOENT / spawnSync /bin/sh ENOENT
```

The `|| true` is inside the shell string and never runs, because the shell is never spawned. So the test throws and **CI goes red**. This file is `no-simulate-phase-in-src.test.ts` — not `*.integration.test.ts` — so it matches the orchestrator vitest config's `src/**/*.test.ts` include and runs in the `ci` job, not the `integration` one. It passes locally for me (212 files / 2922 tests green) only because `/workspaces/generacy` happens to exist here, which is exactly why it slipped through.

**The obvious fix is worse than the bug, and that is the part I most want to flag.** Deleting the `cwd` option makes the test *pass vacuously*. Vitest runs with `cwd` = `packages/orchestrator`, so `packages/orchestrator/src`, `packages/cluster-relay/src` and the rest resolve to paths that do not exist there. `grep` then writes "No such file or directory" to **stderr**, exits 2 — and `|| true` swallows the exit code while `execSync` only captures **stdout**. Output is `''`, `expect(output).toBe('')` passes, and FR-012's boundary check silently guards nothing. A green tick on a test that greps zero files is a worse outcome than the red one, and it is the natural thing for a fixer to reach for.

**Fix — derive the root, and assert the targets exist.**

```ts
import { existsSync } from 'node:fs';
import { dirname, resolve } from 'node:path';
import { fileURLToPath } from 'node:url';

// .../packages/orchestrator/src/__tests__/cockpit-gates → repo root
const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '../../../../..');

const TARGET_DIRS = [
  'packages/orchestrator/src',
  'packages/control-plane/src',
  'packages/cluster-relay/src',
  'packages/generacy/src/cli/commands/cockpit',
  'packages/cockpit/src',
];

it('no SIMULATE_PHASE_* identifiers in shipped code', () => {
  // Guard against the vacuous pass: `|| true` swallows grep's "no such file"
  // exit, and execSync captures only stdout — so a wrong root or a renamed
  // package would make this assert `'' === ''` while grepping nothing.
  for (const dir of TARGET_DIRS) {
    expect(existsSync(resolve(REPO_ROOT, dir)), `grep target missing: ${dir}`).toBe(true);
  }
  const cmd = [
    "grep -rE 'SIMULATE_PHASE_[A-Z]+'",
    ...TARGET_DIRS,
    "--exclude-dir=__tests__ --exclude-dir=tests --exclude='*.test.ts' --exclude='*.spec.ts'",
    '|| true',
  ].join(' ');
  expect(execSync(cmd, { cwd: REPO_ROOT, encoding: 'utf-8' }).trim()).toBe('');
});
```

The `existsSync` loop is the load-bearing half. Without it this test has two silent-pass modes (wrong root, renamed package) and no way to tell them from a real pass — and a boundary check that cannot fail loudly is the same defect class as the `frameId` correlation that shipped inert, which is the thing this whole issue exists to prevent.

**Secondary, same edit.** The `packages/generacy/src/cli/commands/cockpit` entry is narrower than its siblings, which are whole `src` trees. A `SIMULATE_PHASE_*` added anywhere else under `packages/generacy/src` — the query client, the retry helper, `options.ts` — is not caught. Either widen it to `packages/generacy/src`, or state in `contracts/production-code-boundary.md` why the cockpit subtree is the whole surface. My Q5 answer asked for "no simulate-a-bug switch ships in production code" without qualification; a scoped grep that reads as comprehensive is the kind of thing a later reader trusts.

christrudelpw (commented):
**Changes requested — one blocking, and it is environmental rather than a design problem.**

The harness is the right shape and every condition I attached to the clarify answers is honoured, including the ones that were easy to skip:

- **Q1's schema condition.** `fake-peer.ts` now does `GateOpenWireSchema.safeParse(msg.data)` / `GateOutcomeWireSchema.safeParse(msg.data)` at the *payload* level, not just `RelayMessageSchema` at the envelope, and tracks `payloadViolations` so FR-006 can assert zero. `research.md:88` states the reason correctly — the real cloud reads `frameId` off raw frame data, and an envelope-only fake would accept any payload shape the cluster emits, which is how `frameId` shipped inert in the first place.
- **Q1's residual.** The spec header, plan Summary and quickstart all carry it verbatim: *"a real cluster over a real WebSocket against a schema-validated cloud fake … Cluster↔cloud semantic divergence is NOT detectable by this harness."* That is the sentence I most expected to get softened.
- **Q4.** FR-004 asserts exactly one `cluster.cockpit` frame at the peer *and* exactly one `'cockpit gate emitted'` log line per `gateId` across 3 wakes — the existing signal, not an invented one.
- **Q5.** Fault injection is entirely fake-side config (`persistGeneration: false`, `omitRunId: true`), no production flags, and the boundary is enforced by a test rather than a promise.
- **SC-002 attribution is real**, not nominal. FR-005 branches on the cell and asserts the `<pre-phase-a>` sentinel under `phase-A-reverted` versus byte-for-byte round-trip otherwise. That is attribution expressed as a passing assertion about degraded behaviour, which is more useful than a failing test.
- All seven verification items map to an FR, including item 5 over a real socket (FR-006 compares the route's 202 `frameId` against `frame.data.frameId` at the peer).

Verified locally: orchestrator unit suite **212 files / 2922 tests green**; the new `cockpit-gates-runid.integration.test.ts` **18/18**. The 4 integration failures I see are the two `redis-queue-adapter.*` files, which fail identically on `develop` (no local Redis) and are unrelated.

**The blocking finding:** `no-simulate-phase-in-src.test.ts:31` hardcodes `cwd: '/workspaces/generacy'`. That is this devcontainer's path; CI checks out to `/home/runner/work/generacy/generacy`, and `execSync` throws ENOENT on a missing cwd (verified, not assumed). This file runs in the `ci` job, so the job goes red.

What matters more than the bug is the fix. Simply dropping the `cwd` option makes the test **pass while grepping nothing** — vitest's cwd is `packages/orchestrator`, the targets resolve to non-existent paths, grep's complaint goes to stderr, `|| true` swallows the exit code, and `execSync` captures only stdout, so the assertion compares `'' === ''`. A boundary check that silently guards nothing is the exact failure class this issue exists to catch. Full replacement in the inline comment — derive the root from `import.meta.url` and assert each grep target exists before running.

packages/orchestrator/src/__tests__/cockpit-gates/no-simulate-phase-in-src.test.ts
christrudelpw: Addressed in 5211834b — please review, and re-open this thread if it still falls short.

christrudelpw (commented):
**Approved** (COMMENT — GitHub rejects self-approval). CI green: `ci` **and** `integration` both pass.

The `ci` job passing is the direct confirmation that mattered — that is the job the boundary test runs in, on a real CI checkout path (`/home/runner/work/generacy/generacy`), which is exactly where the hardcoded `/workspaces/generacy` would have thrown ENOENT.

The fix took both halves. `REPO_ROOT` now derives from `import.meta.url` (verified: resolves to the repo root, and the test passes when run from `/tmp`, so it is genuinely cwd-independent), and the `existsSync` guard over `TARGET_DIRS` closes the vacuous-pass mode where grep would silently scan nothing. It also widened `packages/generacy/src/cli/commands/cockpit` to `packages/generacy/src`, which was the secondary point — the boundary now covers the whole package rather than one subtree.

Local verification on the fixed branch: orchestrator unit suite **212 files / 2922 tests**, and all three cockpit-gates integration suites together **39/39**.

This closes the composed path. Seven verification items, three revert cells with real per-cell assertions, a schema-validated fake at both envelope and payload level, and — importantly — a spec that says plainly what the harness cannot see rather than implying it covers more than it does.

PR #1082, opened 2026-07-29
feat: #1080 1080-problem-cockpit-gate-list

christrudelpw:
## Summary

Closes #1080

---


christrudelpw (commented):
**Approved** (COMMENT — GitHub rejects self-approval). No findings. CI green: `ci` + `integration`.

All five clarify answers are honoured precisely — Q2=B's one-line schema docblock stating the fact plus a cross-link, Q3=B keeping (a) parity and (b) the `runIdSource` omission while dropping the stale (c), Q4=A naming the extension direction without designing the API, Q5=B paraphrased into the file's compressed voice with `agency#471` and `generacy-cloud#894` kept literal.

**I mutation-tested the new guard rather than trusting it**, since a guard-shaped test that cannot fail is the exact defect this issue exists to close. Adding `...(parsed.data.runId !== undefined ? { runId: parsed.data.runId } : {})` to `listInput`:

```
Tests  1 failed | 17 passed (18)
FAIL  … > handler strips runId before invoking client.listGates (client-seam guard)
```

**Only the client-seam guard failed. The wire-level URL guard passed**, because `buildListUrl` still ignores `runId`. That is empirically the scenario the Q1=A answer predicted — handler drop removed, wire still clean, regression invisible at the URL — and it is why the two guards are not redundant. The comment on the new test says exactly this, and it is now demonstrated rather than asserted.

Both grep gates verified on the branch: SC-001 (`runId requires generation|would 400|would produce a 400|RFC-7807` under `mcp/`) returns **0**, down from 5 on develop; SC-002's identifier tokens return 2. Renaming the old test to `'outbound list URL never carries runId (wire-level guard)'` fixes its own name-vs-assertion mismatch — it claimed the client seam while asserting the wire, which is this issue's defect in miniature.

Also verified: `tsc --noEmit` clean, `parity-gate-list.test.ts` 18/18. One local failure in `parity-context.test.ts` reproduces on `develop` and passes in CI here — environmental, not this PR's.

Issue #23, opened 2026-01-18
Attribution calculation engine


christrudelpw:
## Summary

Implement the attribution calculation engine - the component that determines which layer (baseline, protégé, or human) added value in each decision.

## Parent Epic

#2 - Generacy Core Package

## Dependencies

- generacy-ai/contracts - Attribution and metrics schemas
- generacy-ai/contracts - Three-layer decision model schemas

## Purpose

Attribution is core to Humancy's value proposition. By tracking which layer was right, we can:
- Measure human value objectively
- Demonstrate protégé learning
- Generate marketable metrics
- Justify human-in-the-loop costs

## Attribution Model

| Scenario | Attribution | Interpretation |
|----------|-------------|----------------|
| B = P = H ✓ | All aligned | System was right, no unique contribution |
| B = P ≠ H ✓ | Human unique | Human added value baseline/protégé missed |
| B ≠ P = H ✓ | Protégé value | Human's trained wisdom proved valuable |
| B ≠ P ≠ H ✓ | Both added | Collaboration was key |
| B ✓, P ≠ B | Protégé wrong | Protégé diverged incorrectly |
| P ✓, H ≠ P | Human wrong | Human override was incorrect |

### AttributionCalculator
```typescript
interface AttributionCalculator {
  // Calculate attribution after outcome is known
  calculateAttribution(
    decision: ThreeLayerDecision,
    outcome: DecisionOutcome
  ): Attribution;
  
  // Aggregate attributions into metrics
  aggregateMetrics(
    userId: string,
    attributions: Attribution[],
    period: MetricsPeriod
  ): IndividualMetrics;
}

interface Attribution {
  decisionId: string;
  
  // Which layer was correct
  baselineCorrect: boolean | null;
  protegeCorrect: boolean | null;
  humanCorrect: boolean | null;
  
  // Attribution category
  whoWasRight: 'baseline' | 'protege' | 'human_unique' | 'all_aligned' | 'unknown';
  
  // Value source
  valueSource: 'system' | 'protege_wisdom' | 'human_judgment' | 'collaboration' | 'none';
  
  // Counterfactual analysis
  baselineAlternativeOutcome?: string;
  protegeAlternativeOutcome?: string;
}
```

### Outcome Evaluation
```typescript
interface OutcomeEvaluator {
  // Evaluate if a decision option would have worked
  evaluateOutcome(
    decision: DecisionRequest,
    chosenOption: string,
    actualOutcome: string
[...]

PR #1109, opened 2026-08-18
feat: #1106 1106-summary-answersfilesource-s

### LINKED ISSUE #1106: cockpit doorbell: gate answers silently dropped when gateKey and epicRef differ only by owner/repo letter case — every child-issue gate is lost on multi-repo clusters
## Summary

`AnswersFileSource`'s repo-scope filter compares owner/repo with a raw case-sensitive `!==`. GitHub owner/repo names are case-insensitive, and the gate producers do not agree on casing, so on a multi-repo cluster **every human gate answer for the epic's child issues is silently dropped** and the `/cockpit:auto` doorbell never fires. The operator has to tell the running session by hand that they answered a gate.

This is **not** the cross-repo limitation already documented on `AnswersFileSourceOptions.epicRef`. In the failure below the child issues live in the **same repo as the epic** — only the letter case differs.

## Mechanism

`packages/generacy/src/cli/commands/cockpit/doorbell/answers-file-source.ts:645-653`:

```ts
const gateScope = parseIssueRefFromGateKey(gateLine.gateKey);
if (
  gateScope != null &&
  (gateScope.owner !== this.epicScope.owner ||
    gateScope.repo !== this.epicScope.repo)
) {
  // cross-epic drop
```

`epicScope` comes from `parseEpicRef(options.epicRef)` and `epicRef` is `form.ref` — the CLI argument **verbatim**. There is no case normalization anywhere in the doorbell or the cockpit CLI path (`grep -rn toLowerCase` over `dist/cli/commands/cockpit/` returns only unrelated hits in `plan-dependency-extractor`, `status/group`, `startup-retry`).

## Observed casing rule

Measured against live `answers.ndjson` on two clusters, restricted to generation-suffixed (i.e. `/cockpit:auto`-era) gates:

| Gate family | Ref casing |
|---|---|
| Epic-level (`phase-queue`, `scope-drained`) | epic ref **as typed** → `painworth/doc-intel#3` |
| Child-issue gates in the cluster's **primary** repo | lowercase → matches |
| Child-issue gates in a **secondary** (`repos.dev`) repo | GitHub **canonical** casing → `Painworth/doc-intel#23` |

```
painworth-doc-intel-3-20260818-173346 lower painworth/doc-intel#3  phase-queue           x1
painworth-doc-intel-3-20260818-173346 UPPER Painworth/doc-intel#21 clarification         x1
painworth-doc-intel-3-20260818-173346 UPPER Painworth/doc-intel#22 clarification         x1
painworth-doc-intel-3-20260818-173346 UPPER Painworth/doc-intel#23 clarification         x1
painworth-finetooth-3-20260817-185255 lower painworth/finetooth#14 implementation-review x4
```

Note the epic `painworth/doc-intel#3` and the child `Painworth/doc-intel#23` are the **same repo**, differing only in the `P`.

## Reproduction / impact

Replaying the real 168-line `answers.ndjson` through the actual filter function:

```
epicRef=painworth/doc-intel#3   EMIT=4    DROP=164
epicRef=Painworth/doc-intel#3   EMIT=81   DROP=87
```

Bound as typed (`painworth/...`), **164 of 168** answers are dropped — including the three clarification answers a human had just submitted from the inbox. No single casing works: the other binding rescues the child gates but then drops every `phase-queue` gate.

Cluster A (`repos.dev: [Painworth/doc-intel]`, epic in the secondary repo) reproduces on every gate. Cluster B (`repos.dev` empty) never reproduces — all its refs are uniformly lowercase. That is the entire difference between the two; the doorbell code and the `cockpit-answers-writer` are **byte-identical** across the `stable` and `preview` channels, so upgrading does not help.

## Suggested fix

Compare case-insensitively — GitHub owner/repo are case-insensitive, so this is correct regardless of which producer is "right":

```ts
const sameRepo =
  gateScope.owner.toLowerCase() === this.epicScope.owner.toLowerCase() &&
  gateScope.repo.toLowerCase() === this.epicScope.repo.toLowerCase();
```

The `epicRef` doc comment already proposes "drop the repo filter and rely on downstream `gateId` matching (no open gate ⇒ no match ⇒ harmless over-delivery)" — that option fixes this bug **and** the documented cross-repo limitation in one move, and is worth considering instead.

Separately worth normalizing the producers so `gateKey` casing is stable, but the filter should not depend on that.

## Acceptance

- A gate answer whose `gateKey` differs from the bound `epicRef` only by owner/repo letter case is delivered, not dropped.
- Regression test covering `Painworth/x#1` answer against a `painworth/x#1` epic binding, in both directions.
- Genuine foreign-repo answers are still dropped (or intentionally over-delivered, if the filter is removed).

christrudelpw:
## Summary

Closes #1106

---


packages/generacy/src/cli/commands/cockpit/doorbell/answers-file-source.ts
```diff
@@ -641,12 +641,16 @@ export class AnswersFileSource {
...
+    // because GitHub owner/repo names are case-insensitive and gate producers
+    // disagree on casing (operator-typed vs. lowercase vs. GitHub-canonical);
+    // a raw `!==` here silently drops every same-repo answer that differs only
+    // by letter case.
     const gateScope = parseIssueRefFromGateKey(gateLine.gateKey);
     if (
       gateScope != null &&
-      (gateScope.owner !== this.epicScope.owner ||
-        gateScope.repo !== this.epicScope.repo)
+      (gateScope.owner.toLowerCase() !== this.epicScope.owner.toLowerCase() ||
```
christrudelpw: **BLOCKING — Q2=B audit missed the peer doorbell source.**

`packages/generacy/src/cli/commands/cockpit/doorbell/webhook-to-event.ts` (lines **154, 159, 193, 214**) still matches owner/repo **case-sensitively** via `Set` membership. Same defect class as this fix, same directory, instantiated by the same `doorbell.ts` invocation.

Failure scenario — strictly worse than the bug this PR fixes:
`smee-source.ts` `buildRefSet` (lines 53-83) builds `watchedRepos`/`issues`/`prs` from `resolved.parsed.allRefs`, whose `IssueRef.repo` is captured verbatim from the epic body by `packages/cockpit/src/resolver/ref-shapes.ts` with **no case normalization anywhere in `packages/cockpit/src`**. `webhook-to-event.ts` compares those keys against `repository.owner.login` / `repository.name` from the GitHub payload, which is always GitHub-canonical casing.

So an epic body listing `[#23](https://github.com/painworth/doc-intel/issues/23)` while the canonical repo is `Painworth/doc-intel` gives `watchedRepos={'painworth/doc-intel'}`; the webhook yields `'Painworth/doc-intel'`; `refSet.watchedRepos.has()` is false; **every `issues` / `pull_request` / `check_run` webhook returns null and the smee doorbell never fires at all.**

Fix: normalize both the `refKey`/`repoKey` construction in `webhook-to-event.ts` and the set build in `smee-source.ts`, with matching regression tests in this PR.

packages/generacy/src/cli/commands/cockpit/doorbell/answers-file-source.ts
```diff
@@ -641,12 +641,16 @@ export class AnswersFileSource {
...
+        gateScope.repo.toLowerCase() !== this.epicScope.repo.toLowerCase())
```
christrudelpw: **Non-blocking — second Q2=B instance.**

`packages/generacy/src/cli/commands/cockpit/scope/writer.ts:53` — `lineMatchesRef` compares `parsed.repo === ref.repo` case-sensitively.

Failure scenario: epic body contains `- [ ] Painworth/doc-intel#24` and `cockpit_scope_add` is called with `{repo:'painworth/doc-intel', number:24}` → `bodyContainsRef` returns false, the ref is treated as absent, producing a **duplicate task-list entry** (or a silent no-op on the remove path).

packages/generacy/src/cli/commands/cockpit/doorbell/answers-file-source.ts
```diff
@@ -641,12 +641,16 @@ export class AnswersFileSource {
...
     // target) is emitted; downstream matches on gateId. Foreign-repo answers
-    // are dropped + logged.
+    // are dropped + logged. The owner/repo comparison is case-insensitive
+    // because GitHub owner/repo names are case-insensitive and gate producers
+    // disagree on casing (operator-typed vs. lowercase vs. GitHub-canonical);
+    // a raw `!==` here silently drops every same-repo answer that differs only
+    // by letter case.
     const gateScope = parseIssueRefFromGateKey(gateLine.gateKey);
     if (
       gateScope != null &&
```
christrudelpw: **Non-blocking — third Q2=B instance.**

`packages/generacy/src/cli/commands/cockpit/queue.ts:228` — `classifyRow` and the queue loop compare `ref.repo !== targetRepo` case-sensitively.

Failure scenario: a phase whose refs parsed as `painworth/doc-intel#23` from a URL-form link, queued with `targetRepo` `Painworth/doc-intel` derived from another ref's bare form → every row is pushed with `eligibility {kind:'skip', reason:'cross-repo'}` and **no issue is ever labeled or assigned.**

christrudelpw (commented):
Automated implementation review for #1106.

The Q1=C fix on this file is **correct and well-tested** — both sides lowercased, no remaining case-sensitive path on this predicate, no over-normalization (gateKey/gateId/freeText/filePath and the drop-log line all keep original casing), filter retained, no scope creep. Reverting the two-line change fails 4 tests, so the tests genuinely pin the bug.

Requesting changes on the **Q2=B audit**, which was accepted as part of this issue's clarifications ("audit the doorbell/cockpit consumer path for other case-sensitive owner/repo comparisons, and fix any found **in this same PR** with matching regression tests"). Three sibling instances remain. See the inline threads — they are anchored here because those files are not in this diff.

packages/generacy/src/cli/commands/cockpit/doorbell/answers-file-source.ts
christrudelpw: Addressed in 0ba40b18 — please review, and re-open this thread if it still falls short.

packages/generacy/src/cli/commands/cockpit/doorbell/answers-file-source.ts
christrudelpw: Addressed in 0ba40b18 — please review, and re-open this thread if it still falls short.

packages/generacy/src/cli/commands/cockpit/doorbell/answers-file-source.ts
christrudelpw: Addressed in 0ba40b18 — please review, and re-open this thread if it still falls short.

packages/generacy/src/cli/commands/cockpit/doorbell/__tests__/webhook-to-event.test.ts
```diff
@@ -175,6 +175,68 @@ describe('webhookToStreamEvent — Q1=A mapping table', () => {
...
+    expect(ev.event).toBe('label-change');
+    // Payload casing is preserved in the emitted event (repo/url).
+    expect(ev.repo).toBe('O/R');
+    expect(ev.url).toBe('https://github.com/O/R/issues/42');
+  });
+
+  // The RefSetView invariant (owned by smee-source.ts::buildRefSet) is that
+  // set entries are lowercased. This test pins the query-side normalization
+  // in webhook-to-event.ts against the invariant: any payload casing must hit
+  // a lowercased set entry. See smee-source-reconnect.test.ts for the
```
christrudelpw: **The producer-side half of the blocking fix is pinned by no test.**

Verified by mutation: **reverting all four `.toLowerCase()` calls in `smee-source.ts` `repoRefsToSets`/`buildRefSet` leaves the full unit suite AND the integration suite green.** Every fixture — this file's `refSetOf()`, plus `smee-source.integration.test.ts` and `smee-source-reconnect.test.ts` `FAKE_RESOLVED` — already uses lowercase `'o/r'`, so only the *query* side is exercised. The *build* side is unprotected.

Consequence: a future refactor that drops the set-build normalization silently reintroduces the **total smee-doorbell outage** (every `issues`/`pull_request`/`check_run` webhook returning null) with a fully green CI.

Also: the comment here points the reader at a producer-side invariant assertion in `smee-source-reconnect.test.ts` that **does not exist**.

Suggested fix: add a test whose `resolved.parsed.allRefs` fixture uses **mixed/canonical casing** (e.g. `Painworth/doc-intel`) and asserts the built set contains the lowercased key — so reverting the `buildRefSet` normalization fails. Then correct or drop the stale comment.

(Minor, same file: the second added test is a near-exact duplicate of the first.)

packages/generacy/src/cli/commands/cockpit/doorbell/smee-source.ts
```diff
@@ -59,18 +63,18 @@ function repoRefsToSets(allRefs: IssueRef[]): {
   const prs = new Set<string>();
   const repos = new Set<string>();
   for (const ref of allRefs) {
-    const key = `${ref.repo}#${ref.number}`;
+    const key = `${ref.repo.toLowerCase()}#${ref.number}`;
     issues.add(key);
     prs.add(key);
-    repos.add(ref.repo);
+    repos.add(ref.repo.toLowerCase());
   }
```
christrudelpw: **Newly-reachable case-sensitivity bug in the PrSnapshot cache lookup — please fix in this PR.**

Anchored here because the affected line is outside this diff: **`smee-source.ts:375`**, `this.prev.get(snapshotKey(ev.repo, 'pr', ev.number))`.

That lookup uses the **payload's canonical casing** (`ev.repo` from `buildEvent`), while the cache is keyed with **epic-body casing**: `maybeRefreshAggregate` → `runOnePoll` builds `prev` keys from `resolved.parsed.allRefs` (`poll-loop.ts:93`/`145` → `snapshotKey('Painworth/doc-intel', ...)`).

Failure scenario: epic body lists `Painworth/doc-intel#24`, payload arrives as `painworth/doc-intel` → `prev.get` misses → `pr-checks` and `completed:validate` label-change events are emitted with **`checks` undefined** instead of green/red.

This is the same bug class this PR fixes, and it is **newly reachable**: before this change the event was dropped entirely, so the path was dead code. It is now live and silently degraded.

Suggested fix: normalize the key on both the `snapshotKey` construction in `poll-loop.ts` and the lookup at `smee-source.ts:375`, with a regression test pinning it.

christrudelpw (commented):
Automated re-review of the fixer delta (`0ba40b18`) for #1106.

**The blocking finding is fully fixed.** `refKey`/`repoKey` in `webhook-to-event.ts` lowercase both owner and repo, and `repoRefsToSets`/`buildRefSet` in `smee-source.ts` lowercase all four set-entry constructions — build and lookup match, no one-sided normalization. Findings 2 and 3 (`writer.ts`, `queue.ts`) are correct. **No over-normalization**: `buildEvent`'s `repo`/`url`, `RefSetView.epicRef`/`epicRepo`, `targetRepo` used for gh API calls, branch names and the `contents?ref={branch}` path are all untouched, and `pickTargetRepo` dedups by lowercase while returning first-seen original casing.

Mutation-tested each fix: reverting `webhook-to-event.ts` fails 4 unit + 1 integration test, reverting `writer.ts` fails 4, and reverting `classifyRow` / the row loop / `pickTargetRepo` each independently fails 1-2 queue tests.

Two **non-blocking** issues remain — one surfaced by that mutation testing. See inline threads.

packages/generacy/src/cli/commands/cockpit/doorbell/__tests__/webhook-to-event.test.ts
christrudelpw: Addressed in 1f08893f — please review, and re-open this thread if it still falls short.

packages/generacy/src/cli/commands/cockpit/doorbell/smee-source.ts
christrudelpw: Addressed in 1f08893f — please review, and re-open this thread if it still falls short.

PR #1110, opened 2026-08-18
feat: #1107 1107-summary-implement-phase-s

### LINKED ISSUE #1107: implement phase reports success with zero code: product-diff guard excludes only specs/ and measures the cumulative branch diff, so an earlier phase's CLAUDE.md edit always satisfies it
## Summary

The implement phase's "produced no product-code changes" guard is defeated by a file every speckit branch already carries. An implement phase that writes **zero code** is recorded `phase_complete`, the issue advances to `waiting-for:implementation-review`, and a PR containing only spec artifacts is opened for review. On one repo this has already produced a **merged** PR with 0 source files, whose phantom completion is now the unmet dependency blocking two further issues.

## Mechanism

`packages/orchestrator/src/worker/product-diff.ts:12`:

```ts
export const EXCLUDED_PATH_PREFIXES: readonly string[] = ['specs/'];
```

and `computeProductDiff` (same file, L61-68) diffs `baseRef...HEAD` — the **cumulative branch diff**, not the diff the implement phase itself produced. `phase-loop.ts` then fails the phase only when `productFiles.length === 0`.

Every speckit branch carries a `CLAUDE.md` edit written by an earlier phase (the spec-kit `update_agent` step). So:

```
changedFiles = ["CLAUDE.md", "specs/<slug>/spec.md", ...11 more under specs/]
productFiles = ["CLAUDE.md"]        // non-empty → guard passes
```

The guard can never fire on a speckit branch, whatever the implement phase did.

## Evidence

Issues `Painworth/doc-intel#23` (PR #64) and `#21` (PR #65), same run. The implement commit on each branch:

```
$ gh api repos/Painworth/doc-intel/commits/0e3b5779 --jq '{msg:.commit.message, files:[.files[].filename]}'
{"files":["specs/023-phase-3-port-planner/conversation-log.jsonl"],
 "msg":"chore(speckit): complete implement phase for #23"}
```

`+2` lines, conversation log only. Cumulative branch diff vs `develop`:

```
["CLAUDE.md","specs/023-phase-3-port-planner/clarifications.md", ... ,"specs/023-phase-3-port-planner/tasks.md"]
```

`CLAUDE.md` was added by commit `2e36e21d` (repo scaffolding), long before this branch — it is not the implement phase's work, but it is what satisfies the guard. Both issues nonetheless reached `waiting-for:implementation-review` with `completed:tasks` and every task in `tasks.md` unchecked.

Blast radius across 26 PRs in `Painworth/doc-intel`: exactly three shipped 0 source files — PR #56 (issue #17, **MERGED** spec-only) and PRs #64/#65 (issues #23/#21, open). #56's phantom completion is precisely what now blocks #21: its implement agent verified `origin/develop:packages/provider-port/src/` still contains only the `// Filled by #21` placeholder.

## Why the agent produced nothing

Worth recording, because the guard is the safety net for exactly this: both implement agents hit a **real** blocker and correctly halted to ask the operator a question — a "pre-phase dependency PR" adding runtime deps to `develop` is a manual, out-of-band step with no owner in the speckit pipeline, and it had never been created. Headless there is nobody to answer, the CLI exits 0, and the orchestrator reads exit 0 as success. The agents behaved well; the pipeline recorded their halt as a completed implementation.

## Suggested fix

Two independent defects — **either** alone would have caught this, and both are worth doing:

1. **Exclusion list is incomplete.** `EXCLUDED_PATH_PREFIXES` should also exclude the spec-kit `update_agent` targets, which are agent context, not product code: `CLAUDE.md`, `AGENTS.md`, `GEMINI.md`, `.github/copilot-instructions.md`. Note these are exact filenames, and the current matcher is `startsWith` — a bare `'CLAUDE.md'` prefix entry would also swallow a hypothetical `CLAUDE.md.bak`; prefer an explicit exact-match set alongside the prefix list.

2. **Wrong diff window.** The guard should measure the diff the phase itself produced (e.g. the phase's start commit → `HEAD`) rather than the branch's cumulative diff against the PR base. As written, any product file touched by any earlier phase permanently satisfies the guard for every later phase on that branch.

Consider also failing loudly when the implement phase checks off **no** tasks in `tasks.md` — that is the symptom an operator actually notices, and it is independent of which files changed.

## Acceptance

- An implement phase whose own diff is empty (or lives entirely under excluded paths / agent-context files) fails the phase instead of advancing to `waiting-for:implementation-review`.
- Regression test: branch already carrying a `CLAUDE.md` edit from an earlier phase + an implement phase that writes nothing ⇒ phase fails.
- A phase that legitimately edits `CLAUDE.md` *as its product* is not falsely failed (or this is explicitly accepted).

christrudelpw:
## Summary

Closes #1107

---

### DIFF
--- a/packages/orchestrator/src/worker/phase-loop.ts
+++ b/packages/orchestrator/src/worker/phase-loop.ts
@@ -18,7 +18,8 @@ import { postClarifications, hasPendingClarifications, integrateClarificationAns
 import { PENDING_ANSWER_LITERAL } from '@generacy-ai/workflow-engine';
 import { buildSiblingPromptBlock } from './sibling-prompt.js';
 import { checkSiblingReviews } from './sibling-review-checker.js';
-import { EXCLUDED_PATH_PREFIXES, computeProductDiff, resolveBaseRef } from './product-diff.js';
+import { EXCLUDED_PATH_PREFIXES, EXCLUDED_EXACT_PATHS, computePhaseScopedProductDiff, resolveBaseRef } from './product-diff.js';
+import type { PhaseTracker } from '../types/index.js';
 import { boundOutputTail } from './output-tail.js';
 import { synthesizeOutputTail } from './output-tail-synthesis.js';
 import { performBaseMerge, resolveBaseBranch, type BaseMergeRunner } from './base-merge.js';
@@ -30,6 +31,14 @@ import { randomUUID } from 'node:crypto';
 
 /** Phases that MUST produce file changes to be considered successful. */
 const PHASES_REQUIRING_CHANGES: ReadonlySet<WorkflowPhase> = new Set(['implement']);
+
+/**
+ * TTL for the persisted phase-start commit ref (#1107). 7 days — longer than
+ * the 24h dedup default — so the window survives long gate pauses / fixer
+ * timeouts. Expiry degrades to a re-captured (post-resume) window, never to a
+ * silent pass.
+ */
+const PHASE_START_REF_TTL_SECONDS = 7 * 24 * 60 * 60;
 
 /** Narrow a `WorkflowPhase | string` union to WorkflowPhase. */
 function isWorkflowPhase(value: WorkflowPhase | string): value is WorkflowPhase {
@@ -69,6 +78,14 @@ export interface PhaseLoopDeps {
    * Injected by the worker-mode wiring in server.ts.
    */
   failureFingerprintTracker?: FailureFingerprintTracker;
+  /**
+   * #1107: Redis-backed phase tracker used by the implement-phase product-diff
+   * guard to persist the phase-start commit ref across increments. Optional —
+   * when absent, the guard falls back to a per-invocation `getCurrentCommitSha`
+   * capture (post-resume-only window). Injected by the worker-mode wiring in
+   * server.ts.
+   */
+  phaseTracker?: PhaseTracker;
 }
 
 /**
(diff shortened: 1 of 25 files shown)


packages/orchestrator/src/worker/phase-loop.ts
```diff
@@ -316,6 +333,33 @@ export class PhaseLoop {
...
+      // product-diff guard. Must run AFTER the pre-phase base merge (so
+      // merged-in base files are never inside startRef..HEAD) and BEFORE the
+      // CLI spawn (so HEAD is still the pre-work commit). The ref is persisted
+      // once per phase (persist-once → spans all pre-restart increments, Q5);
+      // resumes reuse the first-entry ref. Only phases that require product
+      // changes need it. Left as `undefined` on any failure — the guard's
+      // try/catch then routes to the product-diff-error detection path (SC-005).
+      let phaseStartRef: string | undefined;
+      const phaseStartRefKey = PHASES_REQUIRING_CHANGES.has(phase)
+        ? `phase-start-ref:${context.item.owner}:${context.item.repo}:${context.item.issueNumber}:${phase}`
```
christrudelpw: **Stale cross-run phase-start-ref key can recreate the fail-open.**

The persisted key is scoped to `owner/repo/issue/phase` only — not to the **branch** or the **run**. A stale ref from a previous failed attempt is silently reused on re-entry.

Failure scenario: implement fails on branch `1107-a` (guard fires; key deliberately left in Redis with `S` = a commit on `1107-a`, TTL 7d). The operator re-queues with `process:speckit-feature`; per the known duplicate-PR-on-re-entry behaviour the run lands on a **new spec-slug branch** `1107-b` cut from a newer develop. Step 2c reads the stale `S` instead of capturing HEAD, so the guard runs `git log --first-parent --no-merges S..HEAD` — a range that now includes the develop commits between the two fork points, i.e. real product files from other people's merges. **A new implement phase that writes nothing at all reports success** — the exact fail-open #1107 set out to close, through a different door.

Same-branch re-queues are milder but still widen the window to the previous attempt's commits.

Suggested fix: include the branch name (and/or the run id) in the key, so a re-entry on a different branch captures a fresh ref.

packages/orchestrator/src/worker/phase-loop.ts
```diff
@@ -316,6 +333,33 @@ export class PhaseLoop {
...
+        : undefined;
+      if (phaseStartRefKey !== undefined) {
+        try {
+          const existing = (await deps.phaseTracker?.getValueRaw(phaseStartRefKey)) ?? null;
+          phaseStartRef = existing ?? (await context.github.getCurrentCommitSha());
```
christrudelpw: **Empty-string phase-start ref is accepted and silently inverts the guard.**

`existing ?? (await getCurrentCommitSha())` only rejects `null`/`undefined`, so an empty-string value — from Redis, or from a `git rev-parse HEAD` that exits 0 with empty stdout — is used verbatim as the start ref.

Failure scenario: `phaseStartRef = ''` → `computePhaseScopedProductDiff` runs `git log --first-parent --no-merges --name-only --pretty=format: ..HEAD`, which git parses as `HEAD..HEAD` (empty left side defaults to HEAD) → exit 0 with an empty file list → `productFiles === []` → **a legitimate implement phase that wrote real code is failed as 'no product-code changes'**, with a `since ` message. Fail-closed rather than fail-open, but a silent false failure.

Neither `GhCliGitHubClient.getCurrentCommitSha` nor the phase-loop asserts a non-empty SHA.

Suggested fix: validate the ref is a non-empty string (ideally SHA-shaped) at both the read and capture sites, and throw rather than proceed on empty.

christrudelpw (commented):
Automated implementation review for #1107.

**Both structural defects are genuinely fixed, and all five accepted clarification decisions are honoured.** `EXCLUDED_EXACT_PATHS` uses root-relative exact equality (Q3=A), the phase-start ref is captured strictly after `performBaseMerge` and traversed with `--first-parent --no-merges` (Q4=A), the ref is persisted in Redis via the new `PhaseTrackerService` raw accessors and wired through `ClaudeCliWorker` so it survives `i--` increments and worker restarts (Q5=B), and there is no escape-hatch marker and no FR-006 logic anywhere in the diff (Q1=A, Q2=A). The tests pin the fix rather than naming it: `SC-002` would pass through without the exact-path set and `SC-004` would pass through without the phase-scoped window.

Requesting changes on two **non-blocking** robustness defects in the new phase-start-ref handling — both in the `@@ -316,6 +333,33 @@` block. See inline threads.

packages/orchestrator/src/worker/phase-loop.ts
christrudelpw: Addressed in d545435d — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/phase-loop.ts
christrudelpw: Addressed in d545435d — please review, and re-open this thread if it still falls short.

PR #1115, opened 2026-08-19
feat: #1113 1113-follow-up-1106-pr

### LINKED ISSUE #1113: cockpit doorbell: pin the PrSnapshot read-through path, not just snapshotKey (follow-up to #1106)
Follow-up to #1106 / PR #1109 (merged as `674cc228`). One non-blocking finding was identified in review and deferred rather than blocking the merge.

## PrSnapshot read-through cache path is untested

`packages/generacy/src/cli/commands/cockpit/doorbell/smee-source.ts:375`

PR #1109 fixed the PrSnapshot cache-key casing bug centrally, inside `snapshotKey` (`watch/snapshot.ts`), and every non-test caller routes through it — `poll-loop.ts:93`/`145` (construction), `aggregate.ts:33/34/43/44`, `smee-source.ts:371` (lookup). That fix is correct and is pinned by the new `snapshotKey — #1106 case-insensitive lookup invariant` tests.

However, those tests pin the **invariant in isolation** (`snapshotKey('A/B', ...) === snapshotKey('a/b', ...)`) rather than exercising the **call path**.

**Failure scenario:** a future refactor inlines the key at `smee-source.ts:375` as `` `${ev.repo}#pr#${ev.number}` `` (payload-canonical casing) while `poll-loop` keeps writing normalized keys. Every `pr-checks` / `completed:validate` event for an epic whose body uses `Painworth/Doc-Intel` is emitted with `checks: undefined` again — and **the whole suite stays green**, because nothing asserts that `processEventBlock` actually routes through `snapshotKey`.

This is the same class of gap that round-2 review caught one level down (the `buildRefSet` normalization was unpinned while looking covered), so it is worth closing properly rather than assuming.

**Suggested fix:** add an integration-level test that drives `processEventBlock` with a `prev` cache populated via `poll-loop`-style construction using mixed-case epic-body refs and a payload carrying canonical casing, asserting the emitted event has `checks` set to the expected green/red value rather than `undefined`.

## Context

The underlying #1106 fix (case-insensitive owner/repo comparison in the doorbell answers-file source, plus the sibling instances in `webhook-to-event.ts`, `smee-source.ts` `buildRefSet`, `scope/writer.ts` and `queue.ts`) is complete and mutation-verified — reverting any individual normalization now fails tests.

christrudelpw:
## Summary

Closes #1113

---


packages/generacy/src/cli/commands/cockpit/doorbell/__tests__/smee-source.integration.test.ts
```diff
@@ -501,6 +501,174 @@ describe('SmeeDoorbellSource integration', () => {
     }, 10_000);
   });
 
+  // #1113 — Pin the read-through PrSnapshot cache path to `snapshotKey`. PR #1109
```
christrudelpw: **The new guard tests never run in the speckit `validate` phase.**

`packages/generacy/vitest.config.ts` declares:

```ts
include: ['src/**/__tests__/**/*.test.ts', '__tests__/**/*.test.ts', 'tests/**/*.test.ts'],
exclude: ['**/*.integration.test.ts'],
```

So `pnpm test` — the first half of the validate command `pnpm test && pnpm build` — **excludes this entire file**. These tests are executed only by CI's separate `integration` job.

Why that matters for *this* issue specifically: #1113 exists to make the `smee-source.ts:375` regression *catchable*. As landed, a future change that inlines the key would sail through `implement` and `validate` reporting success, and only be caught later on the PR by the `integration` job. The protection is real but strictly weaker than it appears, and it is absent from exactly the gate that decides whether an implement phase passed.

**Suggested fix (either is acceptable):**
1. **Preferred** — move the 4 new `it` blocks into a non-integration sibling, e.g. `smee-source.checks-casing.test.ts`, so they match the `include` glob and run in `pnpm test`. They need no live socket or network; the existing `describe` already builds `SmeeDoorbellSource` in-process, so nothing about them requires the integration harness.
2. Alternatively, narrow the `exclude` so this file is included — but note that pulls **every** `*.integration.test.ts` in the package into `validate`, which may be slow or flaky; prefer option 1 unless you want that globally.

Please keep the mutation-sensitivity intact: after the move, re-verify that inlining `smee-source.ts:375` still fails the two read-mixed rows.

christrudelpw (commented):
Automated implementation review for #1113.

**This PR achieves its purpose — verified by mutation.** In a scratch worktree, inlining `smee-source.ts:375` as `` this.prev.get(`${ev.repo}#pr#${ev.number}`) `` took the doorbell integration suite from 19/19 green to **2 FAILED** — exactly the two read-mixed rows (`pr-checks` and `completed:validate`), both `expected undefined to be 'green'`, the real bug's symptom. No pre-existing test noticed the mutation. Separately reverting `.toLowerCase()` inside `snapshotKey` failed all 4 new tests and zero pre-existing ones, confirming the write-mixed rows carry the invariant mutation while the read-mixed rows are the line-375 killers — exactly as the accepted Q1=B rationale predicted.

All three accepted decisions are honoured: **Q1=B** (both directions, load-bearing read-mixed included), **Q2=A** (`+168/-0`, zero deletions — every existing test untouched), **Q3=A** (single `success`→green per branch, both branches covered). The `CLAUDE.md` entry follows the established per-issue convention and its content checks out. No changeset is warranted (`changeset-bot.yml` exempts test-only `packages/*/src/` diffs; the bot passed).

Requesting one change: the new tests are **not reachable from the phase that gates every implement run**. See the inline thread.

packages/generacy/src/cli/commands/cockpit/doorbell/__tests__/smee-source.integration.test.ts
christrudelpw: Addressed in bc8a108d — please review, and re-open this thread if it still falls short.

packages/generacy/src/cli/commands/cockpit/doorbell/__tests__/smee-source.checks-casing.test.ts
```diff
@@ -0,0 +1,343 @@
...
+// `processEventBlock` end-to-end with the write key and read payload in opposite
+// casings; the read-mixed rows go red if `smee-source.ts:375`'s
+// `snapshotKey(ev.repo, 'pr', ev.number)` is ever inlined as
+// `` `${ev.repo}#pr#${ev.number}` `` (payload-canonical casing → cache miss →
+// `checks: undefined`).
+//
+// Lives in a non-`*.integration.test.ts` file so it runs under `pnpm test` and
+// therefore gates the speckit `validate` phase (#1115 review).
+
+interface FakeServer {
```
christrudelpw: **Extract the duplicated test harness into a shared helper.**

The move copied roughly **175 lines of scaffolding verbatim** from `smee-source.integration.test.ts`: `FakeServer` / `startFakeSmee`, `checkRunFrame`, `issueFrame`, `fakePrSnapshot`, `setPrev`, `waitFor`, `FAKE_RESOLVED`, and the `resolveEpic` `vi.mock`. Two independent copies of the same webhook-fake now exist.

Failure scenario: a future fix to the SSE fake server or the webhook frame shape — GitHub changes `check_run.pull_requests`, or `startFakeSmee` needs a keep-alive fix — is applied only in `smee-source.integration.test.ts`. This copy silently drifts, keeps passing against a stale payload shape, and the #1113 cache-key guard stops exercising the real webhook path **while still reporting green**.

That is the same failure mode #1113 exists to prevent: a test that looks like coverage but no longer proves what it claims. Duplicating the harness to fix it is worth closing out properly.

**Suggested fix:** extract the shared scaffolding into a non-test helper module beside these files — e.g. `__tests__/helpers/smee-harness.ts` (any name not matching the `*.test.ts` include glob, so vitest does not try to collect it as a suite) — exporting `startFakeSmee`, `checkRunFrame`, `issueFrame`, `fakePrSnapshot`, `setPrev`, `waitFor` and `FAKE_RESOLVED`, and import it from both files. Keep the `vi.mock('...resolveEpic...')` call in each test file (module mocks are hoisted per-file and do not survive extraction).

Please re-verify after the extraction that inlining `smee-source.ts:375` still fails exactly the two read-mixed rows, and that `vitest list --filesOnly` still collects this file and still does not collect the helper.

(Optional, non-blocking: this file is not Prettier-formatted. generacy CI has no format gate today and the sibling integration file is equally unformatted on develop, so it breaks nothing — but formatting both while touching them would be cheap.)

christrudelpw (commented):
Automated re-review of the fixer delta (`bc8a108d`) for #1113.

**The move worked and the guard survived intact.** Collection is confirmed concretely: this file matches vitest.config.ts's `src/**/__tests__/**/*.test.ts` include and does NOT match the `**/*.integration.test.ts` exclude — `vitest list --filesOnly` lists it among the 202 collected files while `smee-source.integration.test.ts` is absent (0 hits). Both mutations reproduce the pre-move sensitivity exactly: inlining `smee-source.ts:375` as `` this.prev.get(`${ev.repo}#pr#${ev.number}`) `` fails precisely the 2 read-mixed rows (`expected undefined to be 'green'`) with the 2 write-mixed rows still passing, and dropping `.toLowerCase()` from `snapshotKey` (watch/snapshot.ts:47) fails all 4.

The `+343` vs the original `+168` is fully accounted for: a byte-level diff of the four extracted `it` blocks against the ones deleted from the integration file shows **zero differences** — all 4 scenarios survived with identical `toBe('green')` / `toBe('pr-checks')` / `toBe('label-change')` / `toBe('completed:validate')` assertions; nothing weakened, none added or lost. The `+4` left in the integration file is a 3-line pointer comment plus a blank line; no test was stranded and its remaining 15 tests still pass under `vitest.integration.config.ts`. No `.skip`/`.only`/`.todo`, no missing awaits; `tsc --noEmit` and eslint are clean.

Requesting one change before merge — see the inline thread.

packages/generacy/src/cli/commands/cockpit/doorbell/__tests__/smee-source.checks-casing.test.ts
christrudelpw: Addressed in 9cc8830b — please review, and re-open this thread if it still falls short.

PR #1116, opened 2026-08-19
feat: #1112 1112-follow-up-1107-pr

### LINKED ISSUE #1112: orchestrator: phase-start-ref key migration + unresolvable-ref handling (follow-up to #1107)
Follow-up to #1107 / PR #1110 (merged as `d533b41e`). Two non-blocking findings were identified in review and deferred rather than blocking the merge. Both **fail closed** (a false failure, recoverable by operator advance), which is consistent with the Q1=A decision accepted on #1107.

## 1. Key-format change orphans in-flight phase-start refs across the upgrade

`packages/orchestrator/src/worker/phase-loop.ts:366`

#1110 changed the persisted Redis key from `phase-start-ref:<owner>:<repo>:<issue>:<phase>` to `phase-start-ref:<owner>:<repo>:<issue>:<branch>:<phase>`. Refs written by the **currently-deployed** build are never read or cleared by the new build, and linger for the full 7-day TTL.

**Failure scenario (one-time, at rollout):**
1. A cluster on the pre-#1110 build persists `phase-start-ref:owner:repo:1107:implement` = `S` at implement entry.
2. The implement CLI writes all its product code.
3. The worker restarts (npm refresh / `generacy update`) before the phase completes.
4. The new build reads `phase-start-ref:owner:repo:1107:<branch>:implement` → null → re-captures `S'` = current HEAD, **already past the product commits**.
5. The resumed increment writes only spec files → `git log S'..HEAD` yields spec-only paths → a legitimate implement phase is failed as "no product-code changes".

**Suggested fix:** on a miss, read through to the legacy key format once (and clear it), or drain legacy keys at startup.

## 2. Reused persisted ref is shape-validated but never checked to resolve in the current checkout

`packages/orchestrator/src/worker/phase-loop.ts:370`

`isValidCommitSha` confirms the ref is SHA-shaped, but not that the commit exists in this checkout.

**Failure scenario:**
1. Attempt 1 on branch `B`: step 2b `performBaseMerge` commits merge `M` (after `git reset --hard origin/B`); step 2c persists `S = M`.
2. The implement CLI dies (timeout/abort) before step 5 `commitPushAndEnsurePr`, so `M` is **never pushed** and the key is deliberately retained.
3. Re-entry lands on a different worker/container with a fresh clone where `M` does not exist.
4. `M` passes `isValidCommitSha` → `getFilesChangedByOwnCommits(M)` runs `git log --first-parent --no-merges M..HEAD` → exit != 0, `fatal: bad revision` → throw → `product-diff detection failed` + `onError('implement')` escalation — **even though the phase just wrote and pushed real product code at step 5**.

Same-checkout re-entry is unaffected (the walk stops at `M`'s first parent).

**Suggested fix:** verify the reused ref resolves (`git cat-file -e <sha>^{commit}`) before use; on failure, treat as absent and re-capture.

## Not in scope here

FR-006 (zero-tasks-checked net) remains deferred per the Q2=A decision on #1107 and is tracked separately.

christrudelpw:
## Summary

Closes #1112

---


packages/orchestrator/src/worker/__tests__/phase-loop.product-diff.test.ts
```diff
@@ -333,3 +336,140 @@ describe('PhaseLoop - phase-scoped product-diff healthy path (SC-003)', () => {
...
+    await phaseLoop.executeLoop(context, config, deps, ['implement', 'validate']);
+
+    // Shape-invalid legacy → not migrated, but consumed (cleared) exactly once.
+    const legacyClears = clearRaw.mock.calls.filter((c) => c[0] === LEGACY_KEY);
+    expect(legacyClears).toHaveLength(1);
+    // No migration write; fresh HEAD captured instead.
+    expect(setValueRaw).not.toHaveBeenCalledWith(BRANCH_KEY, '', expect.any(Number));
+    expect(context.github.getCurrentCommitSha).toHaveBeenCalled();
+  });
+
```
christrudelpw: **Surviving mutation — Q3 consume-once is unpinned for the shape-valid-but-unresolvable rejection.**

The shipped code is correct: the legacy `clearRaw` at `phase-loop.ts:409` sits outside the accept/reject if-else and before the resolve check, so it fires on all three read outcomes — accepted, shape-invalid, and unresolvable.

But only two of those are pinned. SC-002 covers the **shape-invalid** rejection; there is no test for the **shape-valid-but-unresolvable** one.

Mutation that survives the current suite (18/18 still green): defer the legacy `clearRaw` until after the resolve check and skip it when `commitExistsInCheckout()` returns false.

Failure scenario that would then ship unnoticed: a legacy key holding a well-formed sha for a commit that does not exist in this checkout (the attempt-1-merge-never-pushed case from finding 2 of #1112) survives to its 7-day TTL and is re-read — and re-rejected — on every subsequent branch-scoped miss, with no test failing.

**Suggested fix:** add a sibling to SC-002 that seeds the legacy key with a well-formed sha, stubs `commitExistsInCheckout` to return `false`, and asserts `clearRaw` was still called with the legacy key (and that a fresh HEAD capture followed). That pins the third arm of the consume-once contract and kills the mutation.

christrudelpw (commented):
Automated implementation review for #1112.

**All four accepted answers are correctly implemented, and the Q4 trap was avoided.** The probe is `git rev-parse --verify --quiet <sha>^{commit}` (gh-cli.ts:1431, run with `cwd: this.workdir`), mapping exit 0 -> true, 1 -> false, anything else -> throw with exit code + stderr. Measured on git 2.52.0: existing sha 0; missing 40-hex 1; missing 7-/8-hex 1; blob/tree sha that cannot peel to a commit 1 (safe -> re-capture); outside a repo 128; corrupt `.git/HEAD` 128. The `cat-file -e <sha>^{commit}` probe named in the issue text does return 128 for a missing object *and* for not-a-repo, so it could not have implemented Q4=B — this PR correctly avoided it, and an args assertion pins that choice. The environment-fault throw propagates out of the capture leaving `phaseStartRef === undefined`, which step 5b converts into `product-diff-error` + `onError('implement')` rather than a silent re-capture.

Q1 ordering (setValueRaw at :392 before clearRaw at :409), Q2 (no SCAN/KEYS/scanStream — a single inline GET on a branch-scoped miss), and Q3 (clear sits outside the accept/reject branch and before the resolve check) are all satisfied. Regressions are clean: capture still at step 2c strictly after the 2b `performBaseMerge`, `EXCLUDED_EXACT_PATHS` untouched, `--first-parent --no-merges` unchanged.

Mutation testing killed 6 of 7 mutations (clear-before-write, clear-only-on-accept, resolve-check-removed x2, env-fault-swallowed, `cat-file` substitution, and collapsing the 1-vs-128 exit split). **One survived** — see the inline thread.

Separately, one observation not requested as a change: Q1's "clear only after the branch-scoped write succeeds" is currently *syntactic* ordering only, because `setValueRaw` swallows Redis errors and returns `void`, so the code cannot observe whether the migration write landed. A transient SET failure followed by a successful DEL loses the ref from both keys and reintroduces the #1112 defect on a later restart. Enforcing it needs `setValueRaw` to signal success (or a Lua RENAME / SETNX+DEL), which is an interface change beyond this PR's scope — filing separately rather than blocking here.

packages/orchestrator/src/worker/__tests__/phase-loop.product-diff.test.ts
christrudelpw: Addressed in bae8741e — please review, and re-open this thread if it still falls short.

PR #1150, opened 2026-08-20
feat: #1134 1134-context-bugfix-risk-narrow

### LINKED ISSUE #1134: [orchestrator] Bugfix profiles: verification review charter + targeted validate with diff-classification guards
## Context

Bugfix risk is narrow — wrong root cause, regression in adjacent behavior, missing proof the bug is fixed — and full 10-minute suites per round are unnecessary for one-package fixes. Per-workflow config keys land in generacy-ai/generacy#1122; this issue delivers the profiles that use them.

## Requirements

- Verification review charter (`profile: verification`), four questions: root cause vs symptom; regression test present that fails without the fix; scope beyond what the fix requires; regression risk in changed lines. Defaults for speckit-bugfix: `blockingSeverity: critical`, `maxRemediations: 2`.
- Targeted validate default/guidance for monorepos: `pnpm --filter "...[origin/<base>]" build && pnpm --filter "...[origin/<base>]" test` (changed packages + dependents).
- Engine-side diff classification BEFORE validate, with guards: diff touches root-level config (lockfile, base tsconfig, workspace file, CI workflows) → fall back to the full command; single-package repo → plain command (filter syntax is meaningless there); docs-only diff → skip tests; test-only diff → run those files.
- Opt-in `failThenPass`: run new/changed test files against the base ref and require fail-on-base, pass-on-branch. Off by default.
- Bugfix review on cheaper model/effort exercised via per-workflow agents keying.

## Acceptance

- Unit tests for diff classification and each guard.
- Harness: a bugfix run uses the verification charter, targeted validate, and cap 2; `failThenPass` on/off paths covered.

---
Part of epic generacy-ai/generacy#1120 (engine-native review & remediate phases). Full design: `docs/engine-review-remediate-plan.md` in generacy-ai/tetrad-development; a condensed design summary lives in the epic body.

christrudelpw:
## Summary

Closes #1134

---

### DIFF
--- a/packages/orchestrator/src/worker/fail-then-pass.ts
+++ b/packages/orchestrator/src/worker/fail-then-pass.ts
@@ -1,0 +1,119 @@
+/**
+ * Fail-then-pass regression proof (#1134 US3 / FR-011).
+ *
+ * Opt-in, off by default. When enabled for a `speckit-bugfix` run, this verifies
+ * that the changed test files genuinely exercise the bug: they must FAIL on the
+ * base ref (proving the test reproduces the defect) and PASS on the branch
+ * (proving the fix resolves it). Either direction not holding is a finding.
+ *
+ * The base-ref run happens in a detached git worktree so the branch checkout and
+ * its `node_modules` are never mutated (worktree isolation). The worktree is
+ * always cleaned up in a `finally`, even on error.
+ */
+
+import { execFile } from 'node:child_process';
+import { promisify } from 'node:util';
+import { mkdtemp } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+const execFileAsync = promisify(execFile);
+
+export interface FailThenPassInput {
+  /** The branch checkout root. */
+  checkoutPath: string;
+  /** Base ref to diff/checkout against, `origin/<base>`. */
+  baseRef: string;
+  /** Changed test files (diff set ∩ test globs), repo-relative. */
+  changedTestFiles: string[];
+  /** Abort signal for the spawned test runs. */
+  signal: AbortSignal;
+}
+
+export type FailThenPassResult =
+  | { kind: 'noop' }
+  | { kind: 'pass' }
+  | { kind: 'fail'; reason: 'base-passed' | 'branch-failed'; evidence: string };
+
+interface TestRunOutcome {
+  /** True iff the test run exited 0. */
+  passed: boolean;
+  /** Combined stdout+stderr tail for evidence. */
+  output: string;
+}
+
+/**
+ * Run the given test files with vitest in `cwd`. Never throws — a non-zero exit
+ * (test failure) is captured as `{ passed: false }`.
+ */
+async function runTests(
+  cwd: string,
+  files: string[],
+  signal: AbortSignal,
+): Promise<TestRunOutcome> {
+  try {
+    const { stdout, stderr } = await execFileAsync('pnpm', ['vitest', 'run', ...files], {
+      cwd,
+      signal,
+    });
+    return { passed: true, output: `${stdout}\n${stderr}`.trim() };
+  } catch (err) {
+    const e = err as { stdout?: string; stderr?: string; message?: string };
+    const output = `${e.stdout ?? ''}\n${e.stderr ?? ''}`.trim() || (e.message ?? String(err));
+    return { passed: false, output };
+  }
+}
+
+/**
+ * Verify the changed test files fail on the base ref and pass on the branch.
+ * Empty `changedTestFiles` is a non-blocking `noop`.
+ */
+export async function runFailThenPass(input: FailThenPassInput): Promise<FailThenPassResult> {
+  const { checkoutPath, baseRef, changedTestFiles, signal } = input;
+
+  if (changedTestFiles.length === 0) {
+    return { kind: 'noop' };
+  }
+
+  const worktreePath = join(await mkdtemp(join(tmpdir(), 'gen-ftp-')), 'wt');
+
+  let baseOutcome: TestRunOutcome;
+  try {
+    await execFileAsync('git', ['worktree', 'add', '--detach', worktreePath, baseRef], {
+      cwd: checkoutPath,
+      signal,
+    });
+    baseOutcome = await runTests(worktreePath, changedTestFiles, signal);
+  } finally {
+    // Always remove the worktree, even if the base run threw.
+    await execFileAsync('git', ['worktree', 'remove', '--force', worktreePath], {
+      cwd: checkoutPath,
+      signal,
+    }).catch(() => {
+      /* best-effort cleanup */
+    });
+  }
+
+  if (baseOutcome.passed) {
+    return {
+      kind: 'fail',
+      reason: 'base-passed',
+      evidence:
+        `The changed test file(s) PASSED on the base ref \`${baseRef}\`, so they do not ` +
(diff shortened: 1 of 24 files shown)


packages/orchestrator/src/worker/phase-loop.ts
```diff
@@ -625,12 +648,36 @@ export class PhaseLoop {
...
-            context.signal,
-          );
+          // Validate phase — run test command.
+          // #1134 (US2): for speckit-bugfix, classify the diff and narrow the
+          // built-in default validate command to a pnpm workspace-filter form.
+          // Every other workflow reaches the plain default unchanged (SC-005).
+          let effectiveValidateCommand = config.validateCommand;
+          let targetedValidate: TargetedValidateDecision | undefined;
+          if (context.item.workflowName === 'speckit-bugfix') {
+            targetedValidate = await this.resolveTargetedValidate(context, prManager, config);
```
christrudelpw: **[blocking] Targeted-validate diff resolution has no error handling.** `resolveTargetedValidate()` calls `resolveBaseRef` (network) and `getFilesChangedBetween(baseRef,'HEAD')` (`git diff base...HEAD`, which throws on non-zero exit — e.g. `origin/<base>` not fetched locally, or a transient API error) with no try/catch. The throw propagates to the phase catch (~L708) and hard-fails the speckit-bugfix validate phase. Before this PR, speckit-bugfix validate just ran `config.validateCommand`, so this is a NEW failure surface that is not byte-identical when the diff can't be computed. Wrap resolution in try/catch and fall back to the plain resolved command on any error.

packages/orchestrator/src/worker/fail-then-pass.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  if (changedTestFiles.length === 0) {
+    return { kind: 'noop' };
+  }
+
+  const worktreePath = join(await mkdtemp(join(tmpdir(), 'gen-ftp-')), 'wt');
+
+  let baseOutcome: TestRunOutcome;
+  try {
+    await execFileAsync('git', ['worktree', 'add', '--detach', worktreePath, baseRef], {
```
christrudelpw: **[blocking] Base-ref run has no dependency install, so the regression proof is vacuous.** The fresh `git worktree add --detach` at base has no `node_modules` (not committed, not shared from the branch checkout), so `pnpm vitest run` cannot execute → base always exits non-zero → `baseOutcome.passed=false` unconditionally. Also, newly-added test files do not exist at the base ref → 'no test files found' also fails. Either way base 'fails' for infrastructure reasons, so fail-on-base/pass-on-branch degenerates to 'does the branch pass' and never verifies the regression. Add a `pnpm install` (or share the branch's node_modules) in the worktree, and handle the 'test file absent at base' case explicitly.

packages/orchestrator/src/worker/phase-loop.ts
```diff
@@ -1513,6 +1560,122 @@ export class PhaseLoop {
...
+      context.github,
+      prManager,
+      context.item.owner,
+      context.item.repo,
+    );
+    const base = baseRef.startsWith('origin/') ? baseRef.slice('origin/'.length) : baseRef;
+    const changedFiles = await context.github.getFilesChangedBetween(baseRef, 'HEAD');
+    const isWorkspace = existsSync(join(context.checkoutPath, 'pnpm-workspace.yaml'));
+    const classification = classifyDiff({ changedFiles, isWorkspace });
+    const isBuiltInDefault = config.validateCommand === DEFAULT_VALIDATE_COMMAND;
```
christrudelpw: **[blocking] Built-in-default detection ignores the per-workflow validateCommand tier (Q1=B deviation).** `isBuiltInDefault = config.validateCommand === DEFAULT_VALIDATE_COMMAND` bypasses the resolved per-workflow value. `resolveWorkflowOverrides` resolves `wf?.validateCommand ?? settings?.validateCommand ?? config.validateCommand`, so if an operator sets `settings.workflows['speckit-bugfix'].validateCommand='make check'` while leaving the cluster default, `config.validateCommand` still equals `DEFAULT_VALIDATE_COMMAND` → `isBuiltInDefault=true` → the operator's custom command is silently rewritten to the pnpm filter form. Q1=B requires a custom validateCommand be left untouched. Compare against the RESOLVED per-workflow command, not `config.validateCommand`.

christrudelpw (commented):
**cockpit:auto implementation review — request changes (3 blocking findings)**

Posted as a COMMENT-event review (self-authored cluster PR). See the 3 inline comments:
1. `phase-loop.ts` targeted-validate diff resolution lacks error handling → breaks byte-identical/can't-narrow rollout safety.
2. `fail-then-pass.ts` base-ref worktree has no `pnpm install` → fail-then-pass regression proof is vacuous (feature non-functional).
3. `phase-loop.ts` built-in-default detection ignores the per-workflow `validateCommand` tier → Q1=B deviation.

Please address these in the remediate loop.

packages/orchestrator/src/worker/phase-loop.ts
christrudelpw: Addressed in be3a1af8 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/fail-then-pass.ts
christrudelpw: Addressed in be3a1af8 — please review, and re-open this thread if it still falls short.

packages/orchestrator/src/worker/phase-loop.ts
christrudelpw: Addressed in be3a1af8 — please review, and re-open this thread if it still falls short.

christrudelpw (commented):
**cockpit:auto re-review (round 2) — request changes (1 finding)**

Round-1's 3 findings are all resolved and verified — thank you. One narrow new edge introduced by the fix commit; see the inline comment. Posted as a COMMENT-event review (self-authored cluster PR).

packages/orchestrator/src/worker/fail-then-pass.ts
```diff
@@ -0,0 +1,195 @@
...
+ */
+async function overlayTestFiles(
+  fromRoot: string,
+  toRoot: string,
+  files: string[],
+): Promise<void> {
+  for (const file of files) {
+    const dest = join(toRoot, file);
+    await mkdir(dirname(dest), { recursive: true });
+    await copyFile(join(fromRoot, file), dest);
```
christrudelpw: **[blocking, narrow] `overlayTestFiles` copyFile is unwrapped — re-introduces the infra-error-hard-fails-validate class fixed in round 1.** If `changedTestFiles` includes a test file that was DELETED or RENAMED on the branch (its old path is in the base...HEAD diff but no longer exists in the branch checkout), `copyFile(join(fromRoot, file), dest)` throws ENOENT. The `finally` only removes the worktree (doesn't swallow), so the throw propagates through `runFailThenPass`/`runFailThenPassCheck` to the phase error handler and hard-fails validate on an infrastructure error — exactly what the skip-on-install-failure path was designed to avoid. Trigger is narrow (opt-in failThenPass + a deleted/renamed test file in the bugfix diff), but real. Fix: wrap the overlay copy so a missing source is skipped non-blockingly (mirror the install-failure `skip` mapping) — a test file absent on the branch simply isn't overlaid.

christrudelpw: Addressed in bd66bdc9 — please review, and re-open this thread if it still falls short.

